Issue #10, opened 2022-05-31
Add authentication, profiling to report-api


dinesh-aot:
Reports API in the prototype didn't have proper authentication or profiling in place. This task is to add the authentication and profiling to the reports-api.

PR #11, opened 2022-05-31
add auth and profiling

salabh-aot:
#10
 - added authorization to API end points
 - added profiling to API end points to log the time taken


dinesh-aot (approved):
Looks Good.

PR #44, opened 2022-06-23
fixes + added testcases

SnehaSuresh-aot:


forms-flow-bpm/Dockerfile
```diff
@@ -33,7 +35,7 @@ ENV JAVA_HOME=/opt/java/openjdk \
 
 EXPOSE 8080
 # OpenShift has /app in the image, but it's missing when doing local development - Create it when missing
-RUN test ! -d /app && mkdir /app || :
+# RUN test ! -d /app && mkdir /app || :
```
dinesh-aot, line 38: This should be uncommented.right?

SnehaSuresh-aot: yes

forms-flow-bpm/src/main/java/org/camunda/bpm/extension/hooks/listeners/SyncFormDataPipelineListener.java
```diff
@@ -130,52 +131,52 @@ private void readInnerDataElements(int i, JsonNode jsonData, Set<FormElement> el
...
+		return String.valueOf(execution.getVariables().get("formUrl"));
+	}
+
+	/**
+	 * Returns the endpoint of Sync API.
+	 *
+	 * @return
+	 */
+	private String getSyncApplicationUrl() {
+		return this.reportUrl + "/sync-form-data";
```
dinesh-aot, line 179: Instead of appending strings like this for URLs, can we use something like this.
`import java.nio.file.Paths;
import java.nio.file.Path;
class HelloWorld {
    public static void main(String[] args) {
        Path path = Paths.get("http://localhost", "application", "history");
        System.out.println(path); 
    }
}`

Issue #56, opened 2022-07-06
Implement backend validations for forms for sync-api


dinesh-aot:
**Overivew**
Implement back end validation for the form submission. We can create validation classes for each forms which can be triggered upon calling the sync-api. If this validation fails, the failure will trigger an incident in camunda. we can traced later to analyse what went wrong.

This is primarily to prevent tampering the html or javascript to change values in the front end which is prevented by front end validations.

PR #57, opened 2022-07-07
added validation to sync form data API

salabh-aot:


salabh-aot:
#56 Implement backend validations for forms for sync-api

dinesh-aot (approved):

Issue #108, opened 2022-07-26
Make sure work entries are unique for a given project and work_type


dinesh-aot:
**Overview**
API to be developed to check whether the work being created is unique per project and work type. UI validation can be thrown in case of existing criteria is met. 
Same checking can be done at the backend for validation.

PR #131, opened 2022-08-03
Api to check if work  exists

salabh-aot:
#108 
 added new API to check existing works based on `project_id` and `work_type_id`

### DIFF
--- a/reports-api/src/reports_api/services/work.py
+++ b/reports-api/src/reports_api/services/work.py
@@ -22,6 +22,11 @@ class WorkService:  # pylint: disable=too-few-public-methods
     @classmethod
     def check_existence(cls, project_id, work_type_id):
         """Checks if a work exists for a given project and work type"""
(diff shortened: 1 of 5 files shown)


reports-api/src/reports_api/services/work.py
```diff
@@ -0,0 +1,27 @@
...
+from reports_api.models import Work
+
+
+class WorkService:  # pylint: disable=too-few-public-methods
+    """Service to manage work related operations."""
+
+    @classmethod
+    def check_existence(cls, project_id, work_type_id):
+        """Checks if a work exists for a given project and work type"""
+        if Work.query.filter_by(work_type_id=work_type_id, project_id=project_id).count() > 0:
```
dinesh-aot, line 25: can you add is_deleted check in here.

salabh-aot: updated

dinesh-aot (approved):

PR #187, opened 2022-12-15
Amedments - data updations

salabh-aot:
amendments
 - updated `auto` field value for PECPs
 - added `EVENT` PECP


dinesh-aot (approved):
Looks Good.

PR #593, opened 2023-05-05
Updated the APIs for `staffs, indigenous nations, proponents` to use marshmallow schemas

salabh-aot:
Updated the APIs for `staffs, indigenous nations, proponents` to use marshmallow schemas


reports-api/src/reports_api/resources/indigenous_nation.py
```diff
@@ -15,62 +15,100 @@
...
     def put(indigenous_nation_id):
         """Update and return an indigenous nation."""
-        indigenous_nation = IndigenousNationService.update_indigenous_nation(indigenous_nation_id, API.payload)
-        return indigenous_nation.as_dict(), HTTPStatus.OK
+        indigenous_nation_schema = IndigenousNationSchema()
+        try:
+            request_json = indigenous_nation_schema.load(API.payload)
+        except ValidationError as err:
+            return err.messages, HTTPStatus.BAD_REQUEST
+        exists = IndigenousNationService.check_existence(
```
dinesh-aot: I would prefer to check this condition inside service and throw a custom exception called ResourceExistsError . Then catch it in the resource level and return Conflict http respose.

reports-api/src/reports_api/resources/staff.py
```diff
@@ -94,19 +106,32 @@ class Staff(Resource):
...
     def put(staff_id):
         """Update and return a staff."""
-        staff = StaffService.update_staff(staff_id, API.payload)
-        return staff.as_dict(), HTTPStatus.OK
+        staff_schema = StaffSchema()
+        try:
+            request_json = staff_schema.load(API.payload)
+        except ValidationError as err:
+            return err.messages, HTTPStatus.BAD_REQUEST
+        exists = StaffService.check_existence(
```
dinesh-aot: Existence check of staff should be done based on email address. There can be people with same Firstname and lastname

salabh-aot: Updated

reports-api/src/reports_api/resources/indigenous_nation.py
salabh-aot: Updated

dinesh-aot (approved):

PR #751, opened 2023-06-21
User management APIs

dinesh-aot:
#715


reports-api/src/reports_api/services/user.py
```diff
@@ -0,0 +1,53 @@
...
+        """Get groups that has level set up"""
+        groups = KeycloakService.get_groups()
+        filtered_groups = list(filter((lambda g: 'level' in g['attributes']), groups))
+        return filtered_groups
+
+    @staticmethod
+    def update_user_group(user_id, user_group_request):
+        """Update the group of a user"""
+        existing_group_id = user_group_request['existing_group_id']
+        if existing_group_id is not None and existing_group_id != "":
```
salabh-aot: This could be replaced with `if existing_group_id:` which should account for not None and != ""

Issue #991, opened 2023-09-25
Project Modal - Enhancements


Connor1892:
**As a** user
**I want** to add and make changes to Projects
**so that** I have a database that is up to date

### Scenario 1: 

GIVEN Project tab withing the List Management is open
WHEN I click on 'Create Project' button
THEN I 'Create Project' modal appears with the following fields

- Name (Ghost text "Project Name") (required)
- Proponent (dropdown) (required)
- Type (dropdown) (required)
- Subtype (dropdown) (required)

- Project Description (required)
- Location Description (required) (Ghost text "Provide a detailed description of project's location")

- Latitude (Ghost text "e.g. 22.2222") (number restriction)
- Longitude (Ghost text "e.g. 22.2222") (number restriction)
- ENV Region (dropdown) (required)
- NRS Region (dropdown) (required)
- Capital Investment
- Est. FTE Positions Created (number restricted)
- EPIC GUID
- Certificate Number
- Abbreviation (Ghost text 'EDRMS retrieval code')

- Active (toggle) (default active)
- Closed (toggle) (default inactive)

### Scenario 2:
 
GIVEN Project tab within the List Management is open
WHEN I click on any project name
THEN 'Project Title' modal appears with the following changes to the fields from Scenario 1:

- Name (label) (disabled) with the special history lock button
- Proponent (dropdown) (disabled) with the special history lock button

### Scenario 3:

GIVEN Project modal is open
WHEN I put text value for numerical fields (Latitude, Longtitude, Jobs Created)
THEN I see the error message:

'Please provide numerical values'

### Screenshot 

Create Project
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeUlnQVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--ac84b59f50674b506e28aa3bce8431e14447bdc7/Screen%20Shot%202023-10-10%20at%2010.05.12%20AM.png)

Edit Project
[...]

Issue #990, opened 2023-09-25
Proponent Modal - Enhancements


Connor1892:
**As a** user

**I want** to add and make changes to the Proponent

**so that** I have a database is up to date

### Scenario 1: 

GIVEN 'Proponents' tab within List Management is open 
WHEN I click 'Create Proponent' button
THEN I 'Create Proponent' modal appears with the following information:

- "Create Proponent" Modal Title
- Name (required text field) with the ghost text "Proponent Name"
- Relationship holder (optional dropdown)
- Active toggle (set to 'on' by default)

### Scenario 2:

GIVEN 'Proponents' tab within List Management is open 
WHEN I click on Proponent's Name
THEN I 'Proponent Name' modal appears with the following changes to the fields from Scenario 1:

- 'Proponent name' is the title of the Modal
- Name field is disabled with the lock button

### Screenshot 

Create Proponent
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBdURWIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--eee016235b6903985ced7c830beb34949c838a2b/Screen%20Shot%202023-09-25%20at%205.57.50%20PM.png)

Edit Proponent
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBakhWIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--b410b00325ca62a664b1afcc187f6eae523fc466/image.png)

maria-kuzmenko:
Icons

Lock icon - [Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBdG5WIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--3d5b56e04c4f6ecf9c87db1fff0c68b96f8a1da4/lock-keyhole-bold.svg)

maria-kuzmenko:
link to the prototype

https://www.figma.com/file/j5onJ5RZGAxahxhYp68N8m/Workplan?type=design&node-id=3346-97127&mode=design

Connor1892:
the active toggle is not on by default. When saved, the proponent is active in the grid so this may just be a visual glitch?

[...]

PR #1058, opened 2023-10-16
make active toggle on project and proponent form active by default

TomChapmanGov:
#990 #991


epictrack-web/src/components/proponent/ProponentForm.tsx
```diff
@@ -68,6 +68,10 @@ export default function ProponentForm({ ...props }) {
     reset(ctx.item);
   }, [ctx.item]);
 
+  const setDefaultValues = () => {
+    ctx.setItem({ is_active: true });
```
jadmnsaad: might this overwrite the is_active for an existing item?

TomChapmanGov: if there is an existing item it uses the correct value

dinesh-aot: Would something like this work. This is from FirstNationForm.tsx
`  React.useEffect(() => {
    reset({
      ...workFirstNation,
      work_id: ctx.work?.id,
      is_active: true,
    });
  }, [ctx.work?.id]);`

dinesh-aot (approved):

PR #1071, opened 2023-10-19
Make the workplan filters multi select

jadmnsaad:
task#927

Description:
- Make filters multi select
- Take out event list table to separate component
- Make the loading skeleton for all the workplan 
- Finish loading workphases and add error text if it failed to load
- Add a utility function for filter options for master track table
- Take MasterTrackTable to a separate component


epictrack-api/manage.py
```diff
@@ -24,7 +24,7 @@
 from flask.cli import FlaskGroup
 
 
-APP = create_app(run_mode='migration')
+APP = create_app()
```
dinesh-aot: I think we don't need to remove the 'run_mode'. I know that you did this so that you can run the application.right?

jadmnsaad: I will revert this, I can always just remove it temporarily locally to run it

Issue #1001, opened 2023-09-25
Work Modal - Enhancements


Connor1892:
**As a** user

**I want** to add and make changes to Work

**so that** I have a database that is up to date

### Scenario 1: 

GIVEN 'All Works' tab is open
WHEN I click on 'Create Work' button
THEN 'Create Work' Modal appear with the following information:

- "Create Work" as modal title
- EA Act (Dropdown) (Ghost text "Select EA Act") (required)
- Worktype (Dropdown) (Ghost text "e.g. EAC Assessment") (required)
- Start Date (Date Picker) (Ghost text "MM-DD-YYYY") (required)
- Project (dropdown) (required)
- Responsible Ministry (dropdown) (required)
- Federal Involvement (dropdown) (required)
- Federal Act (dropdown) (required)

- Title (Ghost text "Work Title") (required)
- Project Description (Ghost text "Description will be shown on all reports") (required)
- EPIC Description (Ghost text "Provide a description if differs from the report")
- CAC Required (toggle) (Default inactive) with the following text on hover - 'Select if there is sufficient community interest in this Work to establish a Community Advisory Committee (CAC)'

- EAO Team (dropdown) (required)
- Responsible EPD (dropdown) (required)
- Work Lead (dropdown) (required)
- Decision Maker (dropdown) (required)

- Active (toggle) (default active)
- High Priority (toggle) (default inactive) with the following text on hover - 'Work marked High Priority will have extra milestones appear on Reports'

### Scenario 2:
 
GIVEN 'All Works' tab is open
WHEN I click on 'Edit' button on any Work
THEN 'Project Name - WorkType' Modal appears with the following changes to the current version:

- Modal title is 'Work title' dash 'Worktype'
- Work Title is the combination of 'Work title' - 'Worktype'

- Responsible EPD (dropdown) disabled with the Special History lock button
- Work Lead (dropdown) disabled with the Special History lock button

### Screenshot 

Create Work
![image](https://github.com/bcgov/EPIC.track/assets/126211473/2ced37ff-7656-4b07-98ce-45c0232c735c)

Edit Work
![image](https://github.com/bcgov/EPIC.track/assets/126211473/90bccbec-c91f-4b3b-b55c-3ac80aec242a)

maria-kuzmenko:
Icon

[...]

PR #1079, opened 2023-10-23
Work modal

TomChapmanGov:
#1001


epictrack-api/migrations/versions/eed10a65e3c3_make_columns_not_nullable.py
```diff
@@ -0,0 +1,90 @@
...
+down_revision = '2257493941e7'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    with op.batch_alter_table('works', schema=None) as batch_op:
+        batch_op.alter_column('eao_team_id',
+               existing_type=sa.INTEGER(),
```
dinesh-aot, line 25: I think this will create migration issues as the existing tables has data in dev and test environment. So it is better to truncate the tables before adding these columns.

epictrack-api/src/api/schemas/request/work_request.py
```diff
@@ -26,13 +26,13 @@ class WorkBodyParameterSchema(RequestBodyParameterSchema):
...
-    short_description = fields.Str(
-        metadata={"description": "Short description of work"},
+    report_description = fields.Str(
+        metadata={"description": "Report description of work"},
         validate=validate.Length(max=500),
         required=True,
     )
-    long_description = fields.Str(
-        metadata={"description": "Long description of work"},
+    epic_description = fields.Str(
```
dinesh-aot, line 34: Any particular reason for this name change. Cz the long description would serve the purpose. In the front end we keep changing the label, but it may not be possible to change the backend column name all the time.

epictrack-api/src/api/schemas/request/work_request.py
```diff
@@ -26,13 +26,13 @@ class WorkBodyParameterSchema(RequestBodyParameterSchema):
         required=True,
     )
 
-    short_description = fields.Str(
-        metadata={"description": "Short description of work"},
+    report_description = fields.Str(
```
dinesh-aot, line 29: same as below

epictrack-api/src/api/resources/work.py
```diff
@@ -114,6 +114,7 @@ def get(work_id):
     @profiletime
     def put(work_id):
         """Update and return a work."""
+        print("-----HERE------\n\n")
```
dinesh-aot: Please remove this if this is part of debugging.

epictrack-api/src/api/schemas/request/work_request.py
TomChapmanGov: Only reason for changing the backend column name is so it matched what we have on the front end. Otherwise we just have to remember that report description (frontend) is the short description (backend) and that epic description (frontend) is long description (backend).

TomChapmanGov: But if you want I can revert all the backend changes

PR #1109, opened 2023-10-26
Add Not Applicable option for filters in workplan page

jadmnsaad:
https://app.zenhub.com/workspaces/epictrack-63891ea941d309001fa292cf/issues/gh/bcgov/epic.track/1074
https://app.zenhub.com/workspaces/epictrack-63891ea941d309001fa292cf/issues/gh/bcgov/epic.track/1117

Details:
- Add Blank option to filters options and sort it to be first
- Change highlighted row background color
- Fix background color on select row


epictrack-web/src/components/shared/MasterTrackTable/utils.ts
```diff
@@ -9,7 +10,10 @@ export function getSelectFilterOptions<T>(
   // Step 2: Iterate through the data array to populate the Map
   data.forEach((dataObject) => {
     // Step 3: Skip undefined or null values
-    if (dataObject[key] === undefined || dataObject[key] === null) return;
+    if (dataObject[key] === undefined || dataObject[key] === null) {
+      optionsMap.set("", NOT_APPLICABLE);
```
dinesh-aot: are we supposed to have NA on the dropdowns with no proper values? 
In workplan, I can see NA in end_date which has proper values
responsibilities have two NAs

jadmnsaad: Correct me if I understand the comment wrong, you mean this right:
image
To have NA value when there is no other possible value


In my understanding, NA is also a proper value. And I fear we might over-engineer it if we remove it conditionally. I can check with Maria to get her input on this as well.

dinesh-aot (approved):

PR #1126, opened 2023-10-27
Task modal

TomChapmanGov:
#1016


salabh-aot (approved):

jadmnsaad:
Can we create a folder for the controlled multi select, and all its component would be in it. ControlledMultiSelect folder. For the main component you can name it index.tsx, it will work the same as having ControlledMultiSelect.tsx without the folder

dinesh-aot (approved):

PR #1127, opened 2023-10-27
disable sorting

TomChapmanGov:
#1124


salabh-aot (commented):
If all tables in WorkPlan grid needs this, can we look into disabling at `MasterTrackTable`? Pass a prop for  `enableSorting` which will default to `true` if not provided and use it to disable sort for all columns?

dinesh-aot (approved):

PR #1131, opened 2023-10-30
excluded responsibility ids from task creation

TomChapmanGov:
#1125


epictrack-api/src/api/services/task.py
```diff
@@ -144,6 +144,7 @@ def create_task_events_from_template(
     def _prepare_task_event_object(cls, data: dict) -> dict:
         """Prepare a task event object"""
         exclude = ["assignee_ids"]
+        exclude = ["responsibility_ids"]
```
saravanpa-aot: this is an overwrite ;you sure you dont mean
 exclude = ["assignee_ids", "responsibility_ids"]?

dinesh-aot (approved):

Issue #1106, opened 2023-10-26
Status Tab - Empty State


Connor1892:
**As a** user

**I want** to see when there is no Status associated with the Work

**so that** I can add them

### Scenario 1: 

GIVEN I am on the Status tab of the Workplan
WHEN there is no Status for the Work
THEN I see the following text:

You don't have any Status yet
Create your first Status

'+ Status' (Primary)

### Scenario 2:
 
GIVEN I am in a Work
WHEN there is no status added yet
THEN I see explanation icon next to the status tab

### Scenario 3:
 
GIVEN I am on the Status tab
WHEN I look at the right side
THEN I see the tabbed section:
- Status Preview
- Notes (rich text)

### Scenario 4:
 
GIVEN I am on the Status tab
WHEN I have the Status Preview tab open
THEN I see the following:
- refresh button (ghost)
- Report Preview panel
    - Status
    - preview box (with text: "Your status will appear here" (ghost text))

### Validation Rules? 

(If yes, list here)

### Screenshot 

![image](https://github.com/bcgov/EPIC.track/assets/126211473/c9b803a0-9f77-45c3-b01f-39489d370760)

maria-kuzmenko:
link to prototype

https://www.figma.com/file/yZk63tFMjpuCbYLCBpo7OT/Development-(Issues%2FStatus)?type=design&node-id=3738-34955&mode=design

Connor1892:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/EPICtrack-63891ea941d309001fa292cf/issues/bcgov/EPIC.track/1106?planning-poker) with Zenhub @jadmsaadaot @saravanpa-aot @TomChapmanGov

maria-kuzmenko:
@jadmsaadaot , @TomChapmanGov 

1. we should remove all padding from .jss 5

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeTZiQVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--d7110c7d1dbefc82ec8d6f1d3865ebf80f65909b/padding.png)

maria-kuzmenko:
@jadmsaadaot , @TomChapmanGov 

2. border-radius should be set up to 4px

[...]

PR #1138, opened 2023-10-30
Status empty state

TomChapmanGov:
#1106

### DIFF
--- a/epictrack-web/src/components/workPlan/WorkPlanContext.tsx
+++ b/epictrack-web/src/components/workPlan/WorkPlanContext.tsx
@@ -9,6 +9,7 @@ import {
 } from "../../constants/application-constant";
 import { showNotification } from "../shared/notificationProvider";
 import { WorkFirstNation } from "../../models/firstNation";
+import { Status } from "../../models/status";
 
 interface WorkplanContextProps {
   selectedWorkPhase?: WorkPhase;
@@ -22,6 +23,7 @@ interface WorkplanContextProps {
   setWork: Dispatch<SetStateAction<Work | undefined>>;
   firstNations: WorkFirstNation[];
   setFirstNations: Dispatch<SetStateAction<WorkFirstNation[]>>;
+  status: Status[];
 }
 interface WorkPlanContainerRouteParams extends URLSearchParams {
   work_id: string;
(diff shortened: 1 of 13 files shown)


salabh-aot (approved):

jadmnsaad (approved):

epictrack-web/src/components/workPlan/status/StatusContainer.tsx
```diff
@@ -0,0 +1,144 @@
...
+    padding: "1.5rem 0px 1rem 1rem",
+    minHeight: "0px",
+  },
+});
+
+const StatusContainer = () => {
+  const ctx = React.useContext(WorkplanContext);
+  const classes = useStyle();
+  const [selectedTabIndex, setSelectedTabIndex] = React.useState(0);
+  const status = React.useMemo(() => ctx.status, [ctx.status]);
```
jadmnsaad: I don't think we need a useMemo, we can just pull the status from the context as needed. It's better if we destruct the context as well. Instead of ctx = we do { status } =

epictrack-web/src/components/workPlan/WorkPlanContext.tsx
```diff
@@ -22,6 +23,7 @@ interface WorkplanContextProps {
   setWork: Dispatch<SetStateAction<Work | undefined>>;
   firstNations: WorkFirstNation[];
   setFirstNations: Dispatch<SetStateAction<WorkFirstNation[]>>;
+  status: Status[];
```
jadmnsaad, line 26: will the status data be needed across all tabs. Or only needed in the status tab. If only in the status tab I suggest we create a separate context for the status tab. StatusContext

TomChapmanGov: i am not sure if we will be needing it accross other tabs but we need it on the WorkPlanContainer.tsx for the tab options

epictrack-web/src/components/workPlan/status/StatusView.tsx
```diff
@@ -0,0 +1,46 @@
+import React from "react";
+import NoDataEver from "../../shared/NoDataEver";
+import TrackDialog from "../../shared/TrackDialog";
+import StatusForm from "./StatusForm";
+import { WorkplanContext } from "../WorkPlanContext";
+
+const StatusView = () => {
+  const { status } = React.useContext(WorkplanContext);
+  const [showStatusForm, setShowStatusForm] = React.useState<boolean>(false);
```
jadmnsaad, line 9: these perhaps will be needed in another component, like when clone status is pressed. Perhaps we can have these in the context? But no rush now, just your call or we can move forward on this now

Issue #1129, opened 2023-10-30
Create Project Modal - Automated Project Description


Connor1892:
---
[Proponent] proposes to develop the [Project name], a [project type] which would be located approximately [distance]km from [known near population centre/known near landmark] within the boundaries of [the QQQ Region]. The proposed project is anticipated to produce approximately [production yield] per year of [product], and would include [describe major project components].

---

**As a** team member

**I want** a suggested Project description in the create modal tab

**so that** I can choose not to write one myself

### Scenario 1: 

GIVEN ... (you or your condition)
WHEN ... (what you do)
THEN ... (what you see)

### Scenario 2:
 xxxxxx

### Validation Rules? 

(If yes, list here)

### Screenshot 

(Link above will always be the most up-to-date version.)

PR #1140, opened 2023-10-31
default text for project description

TomChapmanGov:
#1129


jadmnsaad (approved):

epictrack-web/src/components/project/ProjectForm.tsx
```diff
@@ -267,6 +267,9 @@ export default function ProjectForm({ ...props }) {
             <Grid item xs={12}>
               <ETFormLabel required>Project Description</ETFormLabel>
               <TextField
+                defaultValue={
```
dinesh-aot: would it work if we set it in the project.ts model, like the way we set the is_active

export const defaultProject = {
  is_active: true,
};

Issue #1102, opened 2023-10-26
Create action logic


dinesh-aot:
Create action logic for the following
- SetPhasesStatus
- SetEventsStatus
- SetWorkState
- SetProjectStatus
- LockWorkStartDate
- CreateWork

PR #1144, opened 2023-11-01
#1102 - Create action logic

salabh-aot:
Added action handlers for following:
  - Lock work start date
  - set events status
  - set phases status
  - set work state
  - set project status


jadmnsaad:
Can we please add some description? Main bullet points of the changes maybe?

Issue #1107, opened 2023-10-26
Status Tab - First Status


Connor1892:
**As a** team member

**I want** to add a status to my Work

**so that** I can populate Reports

### Scenario 1: 

GIVEN I am on the Status Tab
WHEN I click the Add Status Button
THEN I see the Add Status Modal with the following:
- Title (Add Status)
- Date (label with todays date)
- Description (Text box) (500 character limit) (Required)
- Save (Primary)
- Cancel (Ghost)

![image](https://github.com/bcgov/EPIC.track/assets/126211473/cb640d0e-71e6-420a-8bd4-87dc19054740)

### Scenario 2:

GIVEN Add status modal is open
WHEN I click on save button
THEN I see the status box with the following:
- Date (prepopulated with TODAY's date. For selection, only the following date range is available - from -14 thru +21)
- Tags ("Need Approval")
- Status text
- Approval button
- Edit button

### Scenario 3:

GIVEN A status is not approved
WHEN I press the edit button
THEN I see the Edit Status Modal with the Description active

![image](https://github.com/bcgov/EPIC.track/assets/126211473/3568fc09-424c-4461-af50-dff1c5b55280)

### Scenario 4:

GIVEN Status tab is open
WHEN there is no approved status from the last X days (X is the Admin Freshness number)
THEN a warning appears with the following:

The Work status is out of date
Please provide an updated status.

![image](https://github.com/bcgov/EPIC.track/assets/126211473/761e5439-13ca-479d-aebb-c4cbf0f3d888)

### Scenario 5:
 
GIVEN I am on the Status tab
WHEN I have the Status Preview tab open
THEN I see the following:
- refresh button (ghost)
- Report Preview panel
    - Status
    - preview box (with text: "Once approved, Status will appear here" (ghost text))

![image](https://github.com/bcgov/EPIC.track/assets/126211473/91d557e9-d907-451a-bbe6-0166f50f1a44)

### Validation Rules? 

(If yes, list here)

### Screenshot 

(Link above will always be the most up-to-date version.)

maria-kuzmenko:
link to prototype

https://www.figma.com/file/yZk63tFMjpuCbYLCBpo7OT/Development-(Issues%2FStatus)?type=design&node-id=4702-5076&mode=design

Connor1892:
[...]

Issue #920, opened 2023-09-06
Issues Tab - Empty State


Connor1892:
**As a** user

**I want** to see when there are no issues associated with the Work

**so that** I can add them

### Scenario 1: 

GIVEN I am on the Issues tab of the Workplan
WHEN there is no Issues assigned to the Work
THEN I see the following text:

You don't have any Issues yet
Start adding your Issues

'+ Issue' (Primary)

### Scenario 2:
 
GIVEN I am on the Issues tab
WHEN I select the add Issues button
THEN I see the add Issues modal

### Scenario 3:
 
GIVEN I am on the Issues tab
WHEN I look at the right side
THEN I see the tabbed section:
- Reports Preview
- Notes (rich text)

### Scenario 4:
 
GIVEN I am on the Issues tab
WHEN I have the Report Preview tab open
THEN I see the following:
- 30-60-90 button (Primary)
- Referral Schedule button (ghost)
- Report Preview panel
    - Title (of selected report)
    - sub header (project description)
    - system generated project description
    - [Project Name] is in [Phase Name]
    - sub header (Status)
    - Status (pulled from Status Tab)
    - sub header (Issues)
    - preview box (with text: "Your Issues will appear here" (ghost text))

### Scenario 5:
 
GIVEN I am on the Issues tab
WHEN I have the Report Preview tab open AND select the Referral Schedule button
THEN I see the following:
- 30-60-90 button (Ghost)
- Referral Schedule button (Primary)
- Report Preview panel
    - Title (of selected report)
    - sub header (project description)
    - system generated project description
    - [Project Name] is in [Phase Name]
    - sub header (Status)
    - Status (pulled from Status Tab)
    - sub header (Issues)
    - preview box (with text: "Your Issues will appear here" (ghost text))

Connor1892:
@maria-kuzmenko add pics

maria-kuzmenko:
link to prototype

https://www.figma.com/file/yZk63tFMjpuCbYLCBpo7OT/Development-(Issues%2FStatus)?type=design&node-id=4771-9658&mode=design

Connor1892:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/EPICtrack-63891ea941d309001fa292cf/issues/bcgov/EPIC.track/920?planning-poker) with Zenhub @TomChapmanGov @jadmsaadaot @saravanpa-aot

maria-kuzmenko:
[...]

PR #1155, opened 2023-11-02
Added Model and resouce for work status and issues

saravanpa-aot:
Related
 tickets https://github.com/bcgov/EPIC.track/issues/1107, 
https://github.com/bcgov/EPIC.track/issues/920

- Added the Models for work status and issues
- Added a basic resource and service
-


epictrack-api/src/api/models/work_issue_updates.py
```diff
@@ -0,0 +1,39 @@
...
+
+    __tablename__ = 'work_issue_updates'
+
+    id = Column(Integer, primary_key=True, autoincrement=True)
+    description = Column(String(2000), nullable=False)
+
+    work_issue_id = Column(ForeignKey('work_issues.id'), nullable=False)
+    work_issue = relationship('WorkIssues', back_populates='updates')
+
+    def as_dict(self):  # pylint:disable=arguments-differ
```
dinesh-aot, line 33: Even if you don't put this as_dict() method here, you still be able to get it from the BaseModel

saravanpa-aot: do we even need as_dict? since we use marshamallow , i wasnt sure..just kept it there..

dinesh-aot: > do we even need as_dict? since we use marshamallow , i wasnt sure..just kept it there..

Sometimes I use it when I update stuff. like, status.update(status.as_dict(recursive=False)). I want to pass a dict. yes, I can dump a json using Marshmallow response. I haven't tried that. Lets check the possibility

saravanpa-aot: removed for now..

epictrack-api/migrations/versions/8e4bb9cd046c_issue_and_status.py
```diff
@@ -0,0 +1,178 @@
...
+# revision identifiers, used by Alembic.
+revision = '8e4bb9cd046c'
+down_revision = 'f9bcc9a0564b'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table('work_issues',
```
jadmnsaad, line 21: should table name be plural?

saravanpa-aot: no standard conventions..but easier to go with plural

epictrack-api/src/api/resources/__init__.py
```diff
@@ -105,3 +105,5 @@
 API.add_namespace(RESPONSIBILITY_API, path="/responsibilities")
 API.add_namespace(OUTCOME_CONFIGURATION_API, path="/outcome-configurations")
 API.add_namespace(ACT_SECTION_API, path="/act-sections")
+API.add_namespace(WORK_STATUS_API, path='/work/<int:work_id>/status')
```
dinesh-aot: Do you think it would be good to have plurals in the URL like statuses instead of status

saravanpa-aot: done

saravanpa-aot: changed to statuses

dinesh-aot (approved):

Issue #1134, opened 2023-10-30
Complex Actions


Connor1892:
---
- AddPhase
- SetEventDate
- AddEvent

---

**As a** (type of user)

**I want** (goal)

**so that** (reason/ why/ value)

### Scenario 1: 

GIVEN ... (you or your condition)
WHEN ... (what you do)
THEN ... (what you see)

### Scenario 2:
 xxxxxx

### Validation Rules? 

(If yes, list here)

### Screenshot 

(Link above will always be the most up-to-date version.)

maria-kuzmenko:
@Connor1892, this one Dinesh is saying can be completed - we are now working on bugs associated with this but as separate tickets. 

this one is about actions when the new events should be created.

PR #1166, opened 2023-11-03
#1134 - Complex Actions

salabh-aot:
added action handlers for following:
  - AddEvent
  - AddPhase
  - CreateWork
  - SetEventDate


epictrack-api/src/api/services/work.py
```diff
@@ -674,3 +583,96 @@ def check_work_nation_existence(
         if query.count() > 0:
             return True
         return False
+
+    @classmethod
+    def handle_phase(cls, work_phase, phase_event_templates):  # pylint: disable=too-many-locals
```
saravanpa-aot, line 588: nice work for getting a lot of things done..good job..
one improvement i can suggest is to split this method in multiple smaller methods to improve maintainability ? this is specifically important since we dont have lot of tests ..so whoever is going to maintain this might face a hard time to make changes..
Not for this particular PR ;but great if we can 
1) Write comments ..a lot of explain logic
2) try to split this into multiple methods

epictrack-api/src/api/services/work.py
```diff
@@ -674,3 +583,96 @@ def check_work_nation_existence(
...
+        """Create a new work phase and related events and event configuration entries"""
+        work_phase = WorkPhase.flush(WorkPhase(**work_phase))
+        event_configurations = []
+        for parent_config in list(
+            filter(lambda x: not x["parent_id"], phase_event_templates)
```
saravanpa-aot, line 593: some top level method could be looking like
```
@classmethod
def handle_phase(cls, work_phase, phase_event_templates):
    """Create a new work phase and related events and event configuration entries."""
    work_phase = WorkPhase.flush(WorkPhase(**work_phase))
    event_configurations = cls.create_event_configurations(work_phase, phase_event_templates)
    cls.create_events(work_phase, event_configurations)
```
  Again i might be simplifying it and we might need some real effort to split it .. but on the long run ,it will be worth it..

dinesh-aot (approved):

Issue #1162, opened 2023-11-02
Workplan Page - Filter dupes


Connor1892:
---

the workplan page grid filter considers the bolded milestone dates and the un-bolded task dates as separate entries.
we need them to count as the same day

---

**As a** (type of user)

**I want** (goal)

**so that** (reason/ why/ value)

### Scenario 1: 

GIVEN ... (you or your condition)
WHEN ... (what you do)
THEN ... (what you see)

### Scenario 2:
 xxxxxx

### Validation Rules? 

(If yes, list here)

### Screenshot 

(Link above will always be the most up-to-date version.)

PR #1168, opened 2023-11-03
fixed filter dupes

TomChapmanGov:
#1162

### DIFF
--- a/epictrack-web/src/components/shared/MasterTrackTable/utils.ts
+++ b/epictrack-web/src/components/shared/MasterTrackTable/utils.ts
@@ -16,7 +16,14 @@ export function getSelectFilterOptions<T>(
     }
 
     // Step 4: Populate the Map with unique values and their formatted labels
-    optionsMap.set(dataObject[key], formatLabel(dataObject[key]));
+    if (["start_date", "end_date"].includes(key as string)) {
+      optionsMap.set(
+        formatLabel(dataObject[key]),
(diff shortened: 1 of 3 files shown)


epictrack-web/src/components/shared/MasterTrackTable/utils.ts
```diff
@@ -16,7 +16,14 @@ export function getSelectFilterOptions<T>(
     }
 
     // Step 4: Populate the Map with unique values and their formatted labels
-    optionsMap.set(dataObject[key], formatLabel(dataObject[key]));
+    if (["start_date", "end_date"].includes(key as string)) {
+      optionsMap.set(
+        formatLabel(dataObject[key]),
+        formatLabel(dataObject[key])
```
jadmnsaad: I would suggest adding an optional type prop, and if type === 'date' we do this. date keys could be other than start_date or end_date

jadmnsaad (changes requested):
Adding date type prop

epictrack-web/src/components/shared/MasterTrackTable/utils.ts
jadmnsaad: I'm thinking even we can add a formatValue. Which could be default to () => dataOject[key]. I think this is probably a the better suggestion

dinesh-aot (approved):

jadmnsaad (approved):

PR #1177, opened 2023-11-06
Added endpoints for work status updates

saravanpa-aot:
1. POST endpoint for status creation
2. PUT endpoint for updating a particular status
3. PATCH endpoint for approval flow

### DIFF
--- a/epictrack-api/src/api/resources/work_status.py
+++ b/epictrack-api/src/api/resources/work_status.py
@@ -17,9 +17,10 @@ from http import HTTPStatus
 from flask import jsonify
 from flask_restx import Namespace, Resource, cors
 
+from api.schemas import request as req
 from api.schemas import response as res
 from api.services import WorkStatusService
-from api.utils import auth
+from api.utils import auth, profiletime
 from api.utils.util import cors_preflight
 
 API = Namespace("work-statuses", description="Work Statuses")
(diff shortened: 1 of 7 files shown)


epictrack-api/src/api/resources/work_status.py
```diff
@@ -35,5 +36,56 @@ class WorkStatus(Resource):
...
+    @staticmethod
+    @cors.crossdomain(origin="*")
+    @auth.require
+    @profiletime
+    def put(work_id, status_id):
+        """Update work status"""
+        request_dict = req.WorkStatusParameterSchema().load(API.payload)
+        existing_work_status = WorkStatusService.find_work_status_by_id(work_id, status_id)
+        if existing_work_status is None:
+            return {"message": "Work status not found"}, HTTPStatus.NOT_FOUND
```
dinesh-aot, line 67: You can throw a ResourceNotFoundErr("Work status not found") from here.

epictrack-api/src/api/resources/work_status.py
```diff
@@ -35,5 +36,56 @@ class WorkStatus(Resource):
...
+@cors_preflight("PATCH")
+@API.route("/<int:status_id>/approve", methods=["PATCH", "OPTIONS"])
+class ApproveStatus(Resource):
+    """Endpoint resource to manage approving of work status."""
+
+    @staticmethod
+    @cors.crossdomain(origin="*")
+    @auth.require
+    @profiletime
+    def patch(work_id, status_id):
```
dinesh-aot, line 83: We used to Create a path parameter schema to validate the path parameters. Could you please check if that is required at all

epictrack-api/src/api/services/work_status.py
```diff
@@ -22,6 +26,51 @@ class WorkStatusService:  # pylint: disable=too-many-public-methods
...
+
+    @classmethod
+    def create_work_status(cls, work_id, work_status: Dict):
+        """Creates a work status."""
+        work_status = WorkStatusModel(
+            **work_status,
+            posted_by=TokenInfo.get_username(),
+            work_id=work_id
+        )
+        work_status.save()
```
dinesh-aot, line 47: Does it require to store the result in a variable like work_status = work_status.save() to have the created "id" in the result. I am not sure.

epictrack-api/src/api/resources/work_status.py
saravanpa-aot: will do

epictrack-api/src/api/resources/work_status.py
saravanpa-aot: hm..we can discuss this..

epictrack-api/src/api/services/work_status.py
saravanpa-aot: i dont think so.. but will check the response to be sure..

epictrack-api/src/api/resources/work_status.py
```diff
@@ -35,5 +36,56 @@ class WorkStatus(Resource):
...
+
+    @staticmethod
+    @cors.crossdomain(origin="*")
+    @auth.require
+    @profiletime
+    def put(work_id, status_id):
+        """Update work status"""
+        request_dict = req.WorkStatusParameterSchema().load(API.payload)
+        existing_work_status = WorkStatusService.find_work_status_by_id(work_id, status_id)
+        if existing_work_status is None:
```
jadmnsaad, line 66: do we prefer having this logic at the resource layer rather than the service layer?

epictrack-api/src/api/resources/work_status.py
```diff
@@ -35,5 +36,56 @@ class WorkStatus(Resource):
...
+    """Endpoint resource to manage approving of work status."""
+
+    @staticmethod
+    @cors.crossdomain(origin="*")
+    @auth.require
+    @profiletime
+    def patch(work_id, status_id):
+        """Approve a work status."""
+        existing_work_status = WorkStatusService.find_work_status_by_id(work_id, status_id)
+        if existing_work_status is None:
```
jadmnsaad, line 86: do we prefer having this here rather than the service layer?

PR #1194, opened 2023-11-08
modal active toggle fixes

### LINKED ISSUE #1192: Edit Work Modal - active toggle
---

active toggle isn't changing visually

---

**As a** (type of user)

**I want** (goal)

**so that** (reason/ why/ value)

### Scenario 1: 

GIVEN ... (you or your condition)
WHEN ... (what you do)
THEN ... (what you see)

### Scenario 2:
 xxxxxx

### Validation Rules? 

(If yes, list here)

### Screenshot 

(Link above will always be the most up-to-date version.)

TomChapmanGov:
#1192

### DIFF
--- a/epictrack-web/src/components/indigenousNation/IndigneousNationForm.tsx
+++ b/epictrack-web/src/components/indigenousNation/IndigneousNationForm.tsx
@@ -163,10 +163,7 @@ export default function IndigenousNationForm({ ...props }) {
             <TextField fullWidth {...register("pip_link")} />
           </Grid>
           <Grid item xs={6} sx={{ paddingTop: "30px !important" }}>
-            <ControlledSwitch
-              defaultChecked={(ctx.item as FirstNation)?.is_active}
-              {...register("is_active")}
-            />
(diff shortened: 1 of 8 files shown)


epictrack-web/src/components/indigenousNation/IndigneousNationForm.tsx
```diff
@@ -163,10 +163,7 @@ export default function IndigenousNationForm({ ...props }) {
             <TextField fullWidth {...register("pip_link")} />
           </Grid>
           <Grid item xs={6} sx={{ paddingTop: "30px !important" }}>
-            <ControlledSwitch
-              defaultChecked={(ctx.item as FirstNation)?.is_active}
-              {...register("is_active")}
-            />
+            <ControlledSwitch {...register("is_active")} />
```
jadmnsaad: I think the ...register(..) should be added in the ControlledSwitch itself. like:
image

jadmnsaad (changes requested):
Let's move the register to inside the shared component?

Issue #1220, opened 2023-11-10
Project Notification - Revised PN phases are not created when More information is needed 


maria-kuzmenko:
For Project Notification Work last phase for Decision milestone if a user chooses 'More Information Needed', three phases should be created.

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBL1BHQVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--42bd8b48c8959a4c0e860462d4c608a4f7287aa7/image.png)
 
Currently:
- Only Revised Project Notification Decision phase is created
- Milestones are locked 
- Milestones are unlock when a task is added
- Milestones are locked again and task is deleted when Actual is put in
- Accordion behaviour is tailored to the Project Notification Decision phase 

Desired effect:
- three phases are created with the same Milestones with Anticipated dates only
- accordions for those phases can be opened and closed while pressing on them
- actual dates can be put in and milestone progress would be tracked
- when tasks are added, they stay in the system

maria-kuzmenko:
video with steps reproduced 

[Private Zenhub Video](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBL0hHQVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--ea381bb3583d29eb243368f88329fc3d0f45ad85/20231110150126311.mp4)

carlygreenehill:
- three phases are created with the same Milestones with Anticipated dates only -- YES
- accordions for those phases can be opened and closed while pressing on them -- YES
- actual dates can be put in and milestone progress would be tracked -- YES
- when tasks are added, they stay in the system -- YES

Issue #1171, opened 2023-11-04
Change the way current phase is identified in the work plan


dinesh-aot:
**Problem**
Currently we are keeping current_phase_id in work entity to identify the current phase. Currently a work plan can have duplicated phases, so this method no longer work. We should either store the work_phase_id in work, or just derive the current phase from the work_phase table itself.

**Possible Solutions**
- Keep the current_work_phase_id in work table and update this field whenever a phase is completed
- Derive this information from the work_phase table itself as the table already has a field called is_completed. 
  SELECT * FROM work_phases WHERE is_completed=false order by id desc limit 1
- Create a flag in work_phase called is_current and set it whenever a work phase is completed

maria-kuzmenko:
ways to reproduce: adding actual dates to events or complete phase one by one - next phase should automatically open

PR #1228, opened 2023-11-14
Change the way current phase is identified in the work plan #1171

salabh-aot:
#1171 - change `works.current_phase_id` to `current_work_phase_id`
 #1220 - Project Notification - Revised PN phases are not created when More information is needed


epictrack-api/src/api/schemas/work_v2.py
```diff
@@ -7,13 +7,25 @@
...
 from api.schemas.work_type import WorkTypeSchema
 
 
+class WorkPhaseSchema(
+    AutoSchemaBase
+):  # pylint: disable=too-many-ancestors,too-few-public-methods
+    """Work phase model schema class"""
+
+    class Meta(AutoSchemaBase.Meta):
+        """Meta information"""
```
dinesh-aot, line 22: Why do you have to name it work_v2.py, we could name it work_phase. Plus I always wanted to bring those schemas sitting outside respone or request folder to inside

epictrack-api/src/api/services/work.py
```diff
@@ -119,9 +119,10 @@ def create_work(cls, payload):
             event_template_json = EventTemplateResponseSchema(many=True).dump(
                 event_templates
             )
-            work.current_phase_id = phases[0].id
+            # TODO: CHANGE TO CURRENT WORK PHASE ID
             work = work.flush()
             phase_start_date = work.start_date
+            first_phase = True
```
dinesh-aot, line 125: Can we handle it using for index , phase in enumerate(phases) instead of that extra first_phase variable

epictrack-api/src/api/schemas/work_v2.py
salabh-aot: This is the old schema and I think we are not using this schema anymore. File was probably changed when I did a search for current_phase_id in the entire code base and changed all occurrence to current_work_phase_id

Issue #797, opened 2023-07-10
Rich text enhancement to notes


Connor1892:
---

**As a** User

**I want** to user RTE

**so that** I can leave important messages to my team

### Scenario 1: 

GIVEN Workplan page is open
WHEN Open Notes on First Nations / Issues / Status
THEN I see the following UI changes:

**.rdw-editor-toolbar {**
background: #f9f9fb;
margin-bottom: 8px;
padding: 8px 8px 8px;
border-bottom: 1px solid rgb (224, 224, 224); 

### Screenshot 
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNWV5QVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--6fffaac7eaea92aaa5794bcc61adda9898876c32/Screen%20Shot%202023-11-07%20at%206.42.22%20PM.png)

**element.style {**
height: 485px;
padding: 16px;
resize: vertical;
border: 1px solid rgb(224,224,224);
padding-top: 0px;
border-radius: 4px;

**.public-DraftStyleDefault-block {**
 margin: 8px 0;

### Screenshot 
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNXl5QVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--fa6d44b83c0605dbbe5aeeffcc222fc0f73a4689/Screen%20Shot%202023-11-07%20at%206.46.48%20PM.png)

### Scenario 2: 

GIVEN Workplan page is open
WHEN I click on Notes section
THEN I see the border colour changes to #0070E0

### Screenshot 
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNk95QVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--f289e2ade95da55cf6d629a774fcacaad7397517/Screen%20Shot%202023-11-07%20at%206.59.26%20PM.png)

### Scenario 3: 

GIVEN Workplan page is open
WHEN Open Notes on First Nations / Issues / Status
THEN I see the following changes to functionality:

**Remove:**
[...]

PR #1249, opened 2023-11-15
Rich text

TomChapmanGov:
#797


epictrack-web/src/components/shared/richTextEditor/index.tsx
```diff
@@ -49,6 +77,8 @@ const RichTextEditor = ({
     getStateFromInitialValue();
   }, [initialRawEditorState, initialHTMLText]);
 
+  console.log(focused);
```
jadmnsaad: debug

epictrack-web/src/components/workPlan/event/EventForm.tsx
```diff
@@ -92,6 +92,7 @@ const EventForm = ({
   const [pushEvents, setPushEvents] = useState<boolean>(false);
 
   const ctx = useContext(WorkplanContext);
+  const initialNotes = useMemo(() => event?.notes, [event]);
```
jadmnsaad: [event.id] would probably be safer than [event]. The object might be updated if we did a save. The id will be the same always. Can you do .id for the rest as well?

jadmnsaad (approved):

dinesh-aot (approved):

Issue #1219, opened 2023-11-10
Work states display on Workplans


maria-kuzmenko:
As a user
I want to see what state the Work is in
So that I know what should be done for it

**Work States:**
- Withdrawn
- Closed
- Terminated
- Suspended
- Inactive
- In Progress
- Completed

maria-kuzmenko:
link to prototype - https://www.figma.com/file/j5onJ5RZGAxahxhYp68N8m/For-Development?type=design&node-id=4844-5306&mode=design

PR #1252, opened 2023-11-15
work state display

TomChapmanGov:
#1219

### DIFF
--- a/epictrack-web/src/components/workPlan/WorkState.tsx
+++ b/epictrack-web/src/components/workPlan/WorkState.tsx
@@ -4,13 +4,10 @@ import { Palette } from "../../styles/theme";
 import { Box } from "@mui/system";
 import { WorkplanContext } from "./WorkPlanContext";
 import { Case, Switch } from "react-if";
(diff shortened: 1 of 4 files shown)


epictrack-web/src/components/workPlan/WorkState.tsx
```diff
@@ -0,0 +1,99 @@
...
+import { ETCaption3 } from "../shared";
+import { Palette } from "../../styles/theme";
+import { Box } from "@mui/system";
+import { WorkplanContext } from "./WorkPlanContext";
+import { Case, Switch } from "react-if";
+
+const WorkState = () => {
+  const { work } = useContext(WorkplanContext);
+
+  const capitalizeFirstLetter = (s: string) => {
```
jadmnsaad: Let's avoid using single letter variables.

jadmnsaad: also this looks like it could be a util in a utils or shared file somewhere

dinesh-aot (approved):

Issue #1193, opened 2023-11-08
Resource Forecast - inactive works


Connor1892:
---

inactive works are showing up on the resource forecast

---

**As a** (type of user)

**I want** (goal)

**so that** (reason/ why/ value)

### Scenario 1: 

GIVEN ... (you or your condition)
WHEN ... (what you do)
THEN ... (what you see)

### Scenario 2:
 xxxxxx

### Validation Rules? 

(If yes, list here)

### Screenshot 

(Link above will always be the most up-to-date version.)

dinesh-aot:
**Logic**
is ACTIVE;
WORK Started (ANTICIPATED/ACTUAL in 1st Phase PHASE START EVENT) BEFORE the Report Date + 6 MONTHS!!!!!
WORK did not END (ACTUAL in last "ACTIVE" Phase PHASE END EVENT) BEFORE the Report Date

dinesh-aot:
The downloaded file is corrupted @salabh-aot

PR #1306, opened 2023-11-22
1193 - Resource Forecast - inactive works 

salabh-aot:
#1193 - Resource Forecast - inactive works 
  - hide inactive works from resource forecast report


dinesh-aot:
please revert the package-lock.json to the one from develop and update the PR

jadmnsaad (approved):

dinesh-aot (approved):

Issue #1081, opened 2023-10-24
Automatically add Team Lead to Work Team tab on Work creation


Connor1892:
On team tab change project lead to team lead
On work creation, add work lead as a team member with role team lead

**As a** (type of user)

**I want** (goal)

**so that** (reason/ why/ value)

### Scenario 1: 

GIVEN ... (you or your condition)
WHEN ... (what you do)
THEN ... (what you see)

### Scenario 2:
 xxxxxx

### Validation Rules? 

(If yes, list here)

### Screenshot 

(Link above will always be the most up-to-date version.)

PR #1353, opened 2023-11-24
auto add work lead as team lead

TomChapmanGov:
#1081


epictrack-api/src/api/resources/work.py
```diff
@@ -74,6 +75,8 @@ def post():
         request_json = req.WorkBodyParameterSchema().load(API.payload)
         request_json["start_date"] = get_start_of_day(request_json["start_date"])
         work = WorkService.create_work(request_json)
+        role_id = CodeService.find_code_values_by_type("roles", { "name": "Team Lead" }).get("codes")[0].get("id")
```
dinesh-aot: Can we move this logic to inside create_work function in work.py service. Cz then it will be in a single transaction level with respect to database.

dinesh-aot (approved):

PR #1401, opened 2023-12-04
Generate project abbreviation code for project name

jadmnsaad:
https://app.zenhub.com/workspaces/epictrack-63891ea941d309001fa292cf/issues/gh/bcgov/epic.track/1201

Details:
- Add generate abbreviation endpoint, service and query
- check for duplication
- generate it on blur of project name


epictrack-api/src/api/resources/project.py
```diff
@@ -1,189 +1,205 @@
...
+class ProjectAbbreviation(Resource):
+    """Endpoint resource to import projects."""
+
+    @staticmethod
+    @cors.crossdomain(origin="*")
+    @auth.require
+    @profiletime
+    def post():
+        """Create new project abbreviation"""
+        request_json = request.get_json()
```
salabh-aot: Can we use a schema here? That way we can add validations for required and min lengths and avoid defaulting to `''` in the function call below.

salabh-aot (changes requested):

epictrack-api/src/api/models/project.py
```diff
@@ -1,104 +1,109 @@
...
+            func.lower(Project.name) == func.lower(name), Project.is_deleted.is_(False)
+        )
+        if project_id:
+            query = query.filter(Project.id != project_id)
+        if query.count() > 0:
+            return True
+        return False
+
+    @classmethod
+    def get_by_abbreviation(cls, abbreviation: str):
```
jadmnsaad: added this

epictrack-api/src/api/resources/project.py
```diff
@@ -1,189 +1,205 @@
...
+    @profiletime
+    def post():
+        """Import projects"""
+        file = request.files["file"]
+        response = ProjectService.import_projects(file)
+        return response, HTTPStatus.CREATED
+
+
+@cors_preflight("POST")
+@API.route("/abbreviation", methods=["POST", "OPTIONS"])
```
jadmnsaad, line 193: added this

epictrack-api/src/api/utils/enums.py
```diff
@@ -23,3 +23,11 @@ class HttpMethod(Enum):
     POST = 'POST'
     PATCH = 'PATCH'
     DELETE = 'DELETE'
+
+
+class ProjectCodeMethod(Enum):
```
jadmnsaad, line 28: added this

epictrack-api/src/api/resources/project.py
jadmnsaad: I can add a schema, I'm not sure whether I should add a minimum check. The abbreviation could be custom, which means it can work with any name.

PR #1423, opened 2023-12-07
Card

TomChapmanGov:
- Lazy loading
- Fixed filters expanding when typing
- More responsive view for 1440

### DIFF
--- a/epictrack-web/src/components/myWorkplans/MyWorkplanContainer.tsx
+++ b/epictrack-web/src/components/myWorkplans/MyWorkplanContainer.tsx
@@ -1,21 +1,21 @@
-import { useContext, useRef } from "react";
+import { useContext } from "react";
 import { ETPageContainer } from "../shared";
 import { MyWorkplansContext } from "./MyWorkPlanContext";
 import CardList from "./CardList";
 import { Grid } from "@mui/material";
 import Filters from "./Filters";
 import { AssigneeToggle } from "./Filters/AssigneeToggle";
-import { useOnScreen } from "../../hooks";
+import DummyElement from "../shared/DummyElement";
+import throttle from "lodash/throttle";
 
 const WorkPlanContainer = () => {
   const { loadMoreWorkplans, loadingWorkplans } =
     useContext(MyWorkplansContext);
-  const ref = useRef<HTMLDivElement>(null);
-  const isVisible = useOnScreen(ref);
 
-  if (isVisible && !loadingWorkplans) {
-    setTimeout(() => loadMoreWorkplans(), 250);
-  }
+  const callbackFn = throttle(() => {
+    if (loadingWorkplans) return;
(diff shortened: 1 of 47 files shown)


epictrack-web/src/components/myWorkplans/MyWorkplanContainer.tsx
```diff
@@ -1,32 +1,43 @@
...
 
 const WorkPlanContainer = () => {
-  const {} = useContext(MyWorkplansContext);
+  const { loadMoreWorkplans, loadingWorkplans } =
+    useContext(MyWorkplansContext);
+  const ref = useRef<HTMLDivElement>(null);
+  const isVisible = useOnScreen(ref);
+
+  if (isVisible && !loadingWorkplans) {
+    setTimeout(() => loadMoreWorkplans(), 250);
```
jadmnsaad: Can we create a separate component for the trigger component. And logic like this should be wrapped up in something, a useEffect for instance. https://rasilbaidar.medium.com/trigger-event-when-element-enters-viewport-the-react-way-168509da2e23
image

 Also "setTimeout(() => loadMoreWorkplans(), 250);" let's try doing [throttling](https://www.geeksforgeeks.org/javascript-throttling/) instead of this.

jadmnsaad: We can import throttle from redash like we do with debouncing sometimes:

image
and we can make the throttle interval like 2 seconds

epictrack-web/src/components/workPlan/status/StatusView/index.tsx
```diff
@@ -26,7 +27,11 @@ const StatusView = () => {
           onAddNewClickHandler={() => onAddButtonClickHandler()}
         />
       </When>
-      <When condition={isStatusOutOfDate()}>
+      <When
+        condition={isStatusOutOfDate(
+          statuses.find((status) => status.is_approved)
```
jadmnsaad: can we have a variable defined for all of this const statusOutOfDate = isStatusOutOfDate(......) and have it defined outside the return. in the When it would be just condition={statusOutOfDate}. Same idea for the rest of similar instances

epictrack-web/src/services/workplanService/index.tsx
```diff
@@ -3,8 +3,12 @@ import Endpoints from "../../constants/api-endpoint";
 import { WorkPlan } from "../../models/workplan";
 
 class WorkPlanService {
-  async getAll(page: number) {
-    return await http.GetRequest<WorkPlan[]>(Endpoints.Workplan.GET_ALL);
+  async getAll(page: number, size: number) {
+    console.log("PAGE", page);
```
jadmnsaad: debug

jadmnsaad (approved):

Issue #1332, opened 2023-11-23
Status - Roles


TomChapmanGov:
Refactor status to use new roles logic

**As a** (type of user)

**I want** (goal)

**so that** (reason/ why/ value)

### Scenario 1: 

GIVEN ... (you or your condition)
WHEN ... (what you do)
THEN ... (what you see)

### Scenario 2:
 xxxxxx

### Validation Rules? 

(If yes, list here)

### Screenshot 

(Link above will always be the most up-to-date version.)

ethanjorgensen:
@TomChapmanGov as a viewer, I can still edit approved statuses!

Connor1892:
I can still edit approved status as a viewer

Connor1892:
the edit button for an approved status needs to have the icon be the same disabled colour as the word

Connor1892:
when I changed Maggie in the Admin menu to superuser, she (who is a viewer in keycloak) completely lost access to the system

jadmnsaad:
Is this is now a duplicate to this https://app.zenhub.com/workspaces/epictrack-63891ea941d309001fa292cf/issues/gh/bcgov/epic.track/1586

Connor1892:
I didn't think so, but I'll agree if you say it now is!

Issue #1288, opened 2023-11-20
Admin Menu - Superuser and up


Connor1892:
---

only superusers and higher should be able to see the Admin menu item

---

**As a** (type of user)

**I want** (goal)

**so that** (reason/ why/ value)

### Scenario 1: 

GIVEN ... (you or your condition)
WHEN ... (what you do)
THEN ... (what you see)

### Scenario 2:
 xxxxxx

### Validation Rules? 

(If yes, list here)

### Screenshot 

(Link above will always be the most up-to-date version.)

ethanjorgensen:
@TomChapmanGov I am a viewer, and I can still access the Admin User menu.

Connor1892:
I'm showing up as nothing on Test, but can still see admin menu

Connor1892:
in Dev, I can still see the menu item, but the screen doesn't load data.
once I refreshed it disappeared

PR #1452, opened 2023-12-14
Frontend permission check updates

TomChapmanGov:
#1332 #1288

### DIFF
--- a/epictrack-web/src/components/workPlan/issues/IssueAccordion/Details.tsx
+++ b/epictrack-web/src/components/workPlan/issues/IssueAccordion/Details.tsx
@@ -121,7 +121,10 @@ const IssueDetails = ({ issue }: { issue: WorkIssue }) => {
 
               <Grid item>
                 <If condition={latestUpdate.is_approved}>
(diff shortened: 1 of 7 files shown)


epictrack-web/src/components/workPlan/issues/IssueAccordion/Details.tsx
```diff
@@ -116,20 +120,40 @@ const IssueDetails = ({ issue }: { issue: WorkIssue }) => {
...
-                  }}
-                  onClick={() => {
-                    setUpdateToEdit(latestUpdate);
-                    setShowIssuesForm(true);
-                  }}
-                >
-                  Edit
-                </Button>
+                <If condition={latestUpdate.is_approved}>
+                  <Restricted allowed={[ROLES.EXTENDED_EDIT]}>
```
jadmnsaad: This would hide the wrapped component. You could always pass it props in the  errorProps object. errorProps: {{ disabled: true}}. This would pass the disabled prop to the child

jadmnsaad (approved):

Issue #1359, opened 2023-11-27
Special Field Common Component


dinesh-aot:
**Overview**
Create a common component for special field history

maria-kuzmenko:
@dinesh-aot @Connor1892 

all changes related to Special Filed Common Component are addressed in a separate card #1515

this one can be closed without QA

PR #1461, opened 2023-12-15
Special Field Common Component #1359

salabh-aot:
- Added new shared component for special field
  - bumped material-react-table version to 2.0.5
  - updated project service to add entries to special fields table when creating new project
  - updated special field service to update original table with latest value for the field on create/update operations


dinesh-aot (approved):

epictrack-web/src/components/project/ProjectForm.tsx
```diff
@@ -214,17 +223,20 @@ export default function ProjectForm({ ...props }) {
...
-                      opacity: disabled ? "100" : "0",
-                      cursor: "pointer",
-                    }}
-                    component="img"
-                    src={LockClosed}
-                    alt="Lock"
-                  />
-                </ETFormLabel>
+                <When condition={disabled}>
+                  <If condition={specialField === "name"}>
```
jadmnsaad: please use enums or objects to access these strings, "name", "proponent_id"

epictrack-web/src/components/project/ProjectForm.tsx
```diff
@@ -244,17 +256,22 @@ export default function ProjectForm({ ...props }) {
...
-                    }}
-                    component="img"
-                    src={LockClosed}
-                    alt="Lock"
-                  />
-                </ETFormLabel>
+                <When condition={disabled}>
+                  <If condition={specialField === "proponent_id"}>
+                    <Then>
+                      <IconButton onClick={() => setSpecialField("")}>
```
jadmsaadaot, line 262: if the difference is only in the onClick can we just add a ternary there instead of wrapping the whole button component

epictrack-web/src/components/project/ProjectForm.tsx
```diff
@@ -268,6 +285,54 @@ export default function ProjectForm({ ...props }) {
...
               ></ControlledSelectV2>
             </Grid>
+            <Grid item xs={12}>
+              <When condition={Boolean(specialField)}>
+                <SpecialFieldGrid
+                  entity={SpecialFieldEntityEnum.PROJECT}
+                  entity_id={(ctx.item as Project)?.id}
+                  fieldName={specialField}
+                  fieldLabel={
+                    specialField === "proponent_id" ? "Proponent Name" : "Name"
```
jadmsaadaot, line 295: In this design, what happens if we have more than 2 special fields. Can we refactor this to be more scalable?

dinesh-aot (approved):

PR #1464, opened 2023-12-15
Add cache for filter options and fix name filter text highlight

jadmnsaad:
https://app.zenhub.com/workspaces/epictrack-63891ea941d309001fa292cf/issues/gh/bcgov/epic.track/1463

Details:
- Add cache for project type and work type
- Fix wrapping for text highlight in name filter

### DIFF
--- a/epictrack-api/src/api/resources/project.py
+++ b/epictrack-api/src/api/resources/project.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 """Resource for project endpoints."""
 from http import HTTPStatus
+from api.utils.caching import AppCache
 
 from flask import jsonify, request
 from flask_restx import Namespace, Resource, cors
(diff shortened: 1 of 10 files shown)


epictrack-api/src/api/resources/project.py
```diff
@@ -223,6 +224,7 @@ class ProjectTypes(Resource):
     @cors.crossdomain(origin="*")
     @auth.require
     @profiletime
+    @AppCache.cache.cached(timeout=constants.CACHE_DAY_TIMEOUT, query_string=True)
     def get():
         """Return all project types."""
         project_types = ProjectService.find_all_project_types()
```
dinesh-aot, line 230: This is actualy a GET for the resource TYPE. Shall we move it to it's own resource file

epictrack-api/src/api/resources/work.py
```diff
@@ -470,6 +471,7 @@ class WorkTypes(Resource):
     @staticmethod
     @cors.crossdomain(origin="*")
     @auth.require
+    @AppCache.cache.cached(timeout=constants.CACHE_DAY_TIMEOUT)
```
dinesh-aot, line 474: Same as that of above. This can be a separate resource called WorkType

dinesh-aot (approved):

Issue #1370, opened 2023-11-30
Refactor 30-6-90/EA Referral Reports


dinesh-aot:

dinesh-aot:
**Anticipated Report Logic**
IF WORK ACTIVE = TRUE								AND
IF WORK STATE = In Progress  OR Suspended				AND
IF any (EventType = 5) REFERRAL MILESTONE DATE(anti/actu) equal to or greater than [REPORT DATE - 7]	THEN
Display on Anticipated EA Schedule


**30-60-90**
IF WORK ACTIVE = TRUE										AND
IF WORK STATE = In Progress OR Suspended						AND
								AND
IF any Minister/CEAO (EventTypes = 14, 15)  DECISION MILESTONE DATE, PCP, EVENT HIGH PRIORITY if the work is high priority(anti/actual) is equal to or greater than [REPORT DATE]	AND
IF any Minister/CEAO DECISION MILESTONE DATE, PCP, EVENT HIGH PRIORITY if the work is high priority(anti/actual) is less than [REPORT DATE] + 90		THEN
Display on 30-60-90

PR #1494, opened 2023-12-20
reports refactor

salabh-aot:
-  #1193 
  -  #1370 
    - Fixed date format issue with reports

### DIFF
--- a/epictrack-api/src/api/reports/resource_forecast_report.py
+++ b/epictrack-api/src/api/reports/resource_forecast_report.py
@@ -13,7 +13,7 @@ from reportlab.lib.styles import getSampleStyleSheet
 from reportlab.lib.units import inch
 from reportlab.pdfgen.canvas import Canvas
 from reportlab.platypus import Paragraph, Table, TableStyle
-from sqlalchemy import Date, and_, func
+from sqlalchemy import and_, func, select
 from sqlalchemy.dialects.postgresql import DATERANGE
 from sqlalchemy.orm import aliased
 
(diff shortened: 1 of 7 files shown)


epictrack-api/src/api/reports/resource_forecast_report.py
```diff
@@ -243,6 +243,48 @@ def _fetch_data(self, report_date: datetime):  # pylint: disable=too-many-locals
             .subquery()
         )
 
+        works_started = (
```
saravanpa-aot, line 246: curious , why not use 
 db.session.query(wp_alias.work_id)
same like use filter instead of where?
any reason to not to go for sqlalchemy?
i might be totally wrong as well..

epictrack-api/src/api/reports/thirty_sixty_ninety_report.py
```diff
@@ -47,20 +47,19 @@ def __init__(self, filters):
...
+                EventConfiguration.event_category_id == EventCategoryEnum.PCP.value,
+            )
+        ).scalars().all()
 
-        pecp_category = EventCategory.query.filter(EventCategory.name == "PCP").first()
-        pecp_configuration_ids = EventConfiguration.query.filter(
-            EventConfiguration.event_category_id == pecp_category.id
-        ).all()
-        self.pecp_configuration_ids = [x.id for x in pecp_configuration_ids]
+        self.decision_configuration_ids = db.session.execute(
```
saravanpa-aot, line 57: same as
same as  self.decision_configuration_ids = (
    db.session.query(EventConfiguration.id)
    .filter(EventConfiguration.event_category_id == EventCategoryEnum.DECISION.value)
    .all()
)?

saravanpa-aot (commented):
just a few comment about using sql instead of sqlalchemy..

epictrack-api/src/api/reports/resource_forecast_report.py
salabh-aot: `query` object has no `scalars` method. Without it we will be getting a list of tuples like, `[(id1,), (id2, )...]` and will need to do a list comprehension on it to get a list of ids, (`[x[0] for x in query_result]`). This way, we get a proper list of ids, like, `[id1, id2, ...]` as a result of the db operation itself. **(At least I am not aware of any methods to convert query object into a list of ids without additional actions after db query)**

saravanpa-aot: thanks @salabh-aot for the reply...
yeah..i thought so...
dont think we have a scalar for this purpose..the only way is do a list comprehension after the tuple..
we will have to do a careful analysis when we opt for plain queries..when it grows bigger , we will loose the flexibility of the sqlalchmey..  probably we can do a list comprehension to do a tuple to list... 
But i this case ,query is very concise , so its all good...

saravanpa-aot (approved):

dinesh-aot (approved):

PR #1499, opened 2023-12-21
Add special history component for epd and work lead

jadmnsaad:
https://app.zenhub.com/workspaces/epictrack-63891ea941d309001fa292cf/issues/gh/bcgov/epic.track/1497

https://app.zenhub.com/workspaces/epictrack-63891ea941d309001fa292cf/issues/gh/bcgov/epic.track/1000

Details:
- add a reusable component for the special field lock
- Add separate work form epd special field
- Add separate work form team lead

I aimed not to clutter the work form more


epictrack-api/src/api/services/work.py
```diff
@@ -244,6 +246,28 @@ def create_work(cls, payload, commit: bool = True):
             db.session.commit()
         return work
 
+    @classmethod
+    def create_special_fields(cls, work: Work):
+        """Create work special fields"""
+        work.flush()
+        work_epd_special_field_data = {
```
dinesh-aot, line 254: You still have to call this method in the create_work service method?

epictrack-web/src/constants/application-constant.ts
```diff
@@ -96,9 +96,13 @@ export enum SpecialFieldEntityEnum {
   WORK = "WORK",
 }
 
-export const SpecialFields = {
+export const SPECIAL_FIELDS = Object.freeze({
```
TomChapmanGov, line 100: Just curios, what is the beenfit of doing Object.freeze when its a const?

epictrack-api/src/api/services/work.py
jadmnsaad: Done

epictrack-web/src/constants/application-constant.ts
jadmnsaad: I think it's just good practice. Object.freeze makes the object immutable, it's better practice maybe to use libraries like immutable to deep freeze it.  It's because this is a constant and shouldn't be changed at run time:

In the examples below: SPECIAL_FIELDS is frozen, SPECIAL_FIELDS2 is not frozen:

```
// Attempting to modify a property of the frozen object
try {
  SPECIAL_FIELDS.PROJECT.NAME = "new_name"; // This will throw an error
} catch (error) {
  console.error("Error modifying frozen object:", error.message);
}

// Attempting to add a new property to the frozen object
try {
  SPECIAL_FIELDS.NEW_FIELD = "new_value"; // This will throw an error
} catch (error) {
  console.error("Error adding property to frozen object:", error.message);
}

// Attempting to delete a property from the frozen object
try {
  delete SPECIAL_FIELDS.WORK.RESPONSIBLE_EPD; // This will throw an error
} catch (error) {
  console.error("Error deleting property from frozen object:", error.message);
}

// Modifying a property of the unfrozen object
SPECIAL_FIELDS2.PROJECT.NAME = "new_name"; // This will work

// Adding a new property to the unfrozen object
SPECIAL_FIELDS2.NEW_FIELD = "new_value"; // This will work

// Deleting a property from the unfrozen object
delete SPECIAL_FIELDS2.WORK.RESPONSIBLE_EPD; // This will work
```

Issue #1492, opened 2023-12-19
About Tab - Team Member


Connor1892:
**As a** Team Member

**I want** a place for all the information about a Work and attached Project 

**so that** I am able to see all the information related to my Work in one place

### Scenario 1: 

GIVEN I am on the Work Page for an Assessment or Earlier Worktype
WHEN I click on the About tab
THEN I see the following:
 - Details (default)
 - Calendar (not active. to be added later)

on Details:
- Project panel
- Work panel
- Resources

#### Project Panel:
- Project Creation Date
- Proponent
- Project Description (ability to Read More/Read Less)
- Type
- Subtype
- Location Description (ability to Read More/Read Less)
- ENV Region
- NRS Region
- Abbreviation

#### Work Panel:
- Work Start Date
- Anticipated Referral Date
- Current Milestone (truncated)
- Next Milestone (truncated)
- Work Description (ability to Read More/Read Less)
- EA Act
- Federal Involvement
- Federal Act
- Responsible Ministry
- Decision Makers (title of decision maker, not the name)

<img width="1424" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBM1o2QWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--17880c8564c42129ef8aeed74a3310edaf9b1274/Screen%20Shot%202023-12-20%20at%2012.46.11%20PM.png" alt="Screen Shot 2023-12-20 at 12.46.11 PM.png" />

### Scenario 2:
GIVEN I am on the Work Page for a Post Assessment Worktype
WHEN I click on the About tab
THEN I see the following:

#### Project Panel:
- Certificate Number
- Certificate Issued

![image](https://github.com/bcgov/EPIC.track/assets/126211473/0f831013-8986-410a-864f-95ecfa0ebf03)

### Scenario 3:
GIVEN I am on the About tab
WHEN I look at the Resources panel
THEN I see the following:
- Link to EPIC.Public (contextual to jump to the page about this Work)
- Link to eGuide
- EA Act (2018) (contextual to act this work is under)

#### Project Panel:
- Certificate Number
- Certificate Issued

maria-kuzmenko:
link to prototype - https://www.figma.com/file/E92XoEegK2fLQ3odMeyWBD/Development-(Issues_Status_Dashboard)?type=design&node-id=28-5340&mode=design

[...]

PR #1508, opened 2023-12-28
about page v1

TomChapmanGov:
#1492

### DIFF
--- a/epictrack-web/src/components/workPlan/about/aboutDetails/WorkDetails.tsx
+++ b/epictrack-web/src/components/workPlan/about/aboutDetails/WorkDetails.tsx
@@ -1,0 +1,133 @@
+import { Box, Divider, Grid } from "@mui/material";
+import { useContext, useEffect, useState } from "react";
+import { WorkplanContext } from "../../WorkPlanContext";
+import { ETCaption1, ETCaption2, GrayBox } from "../../../shared";
+import { Palette } from "../../../../styles/theme";
+import dayjs from "dayjs";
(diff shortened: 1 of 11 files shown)


epictrack-web/src/components/workPlan/about/aboutDetails/WorkDetails.tsx
```diff
@@ -0,0 +1,133 @@
...
+import { ETCaption1, ETCaption2, GrayBox } from "../../../shared";
+import { Palette } from "../../../../styles/theme";
+import dayjs from "dayjs";
+import { MONTH_DAY_YEAR } from "../../../../constants/application-constant";
+
+const WorkDetails = () => {
+  const [phaseInfo, setPhaseInfo] = useState<any>({});
+  const { work, workPhases } = useContext(WorkplanContext);
+
+  const getCurrentAndNextPhaseInfo = () => {
```
jadmnsaad: all good, just FYI, you can do something like const [phaseInfo, setPhaseInfo] = useState<any>(getCurrentAndNextPhaseInfo); without having to call it in a useEffect

TomChapmanGov: I think that would be out of scope though?

jadmnsaad: it's okay, just an FYI

jadmnsaad (approved):

Issue #1536, opened 2024-01-03
Resource forecast download bug


dinesh-aot:
**How to reproduce**
- Put any date which has a valid report to show
- Press the download button on top right
- Check the downloaded file by opening it

**Expected**
- The file should show the content as you see in the web

**Actual**
- The file cannot be opened

Connor1892:
@dinesh-aot this needs an estimate!

Issue #1523, opened 2024-01-03
Integration tests for the basic resources


dinesh-aot:
**Overview**
Integration tests for the below mentioned resources
- ActSection
- Code
- EaoTeam
- Outcome Configuration
- Outcome
- Phase
- Position
- ProjectType
- Region
- SubTypes
- WorkType
- Responsibility

PR #1556, opened 2024-01-05
#1523 - added test cases for following APIs (#1547)

saravanpa-aot:
- Act section
  - EAO Teams
  - Positions
  - Project types
  - Regions
  - Responsibility
  - Work type #1536 - Fixed issue with reports download API due to date format


saravanpa-aot:
- [ ]

Issue #1623, opened 2024-01-09
Implement Create/Update/Delete on Import endpoint of Project


dinesh-aot:
**Overview**
- Currently we have just CREATE operation on the import endpoint of Project Resource
- We should have UPDATE/DELETE also available
- Shall we extend python SET with a uniqueness condition to see if the entity exists

Issue #999, opened 2023-09-25
Special History - Proponent name


Connor1892:
**As a** user
**I want** change the history of Proponent's name
**so that** I can track it

### Scenario 1: 

GIVEN Proponent Tab within List Management is open
WHEN I click on any Proponent's Name
THEN I see the modal window with the following changes to the fields:
- Name greyed out with the lock symbol to the right

### Screenshot for Scenario 1:
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBa1RkIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--336525e5127dac43603b78294c53409293ed5be2/Screen%20Shot%202023-09-26%20at%204.01.03%20PM.png)

### Scenario 2:
GIVEN Proponent Edit Modal Window is open
WHEN I click the Lock button next to the Name field
THEN lock button changes to unlock button
THEN the Proponent Name History modal opens with the following information:
- 'Proponent Name' title
- Explanation text - 'Update the legal name of the Proponent and the dates each name was in legal use. Click this link for detailed instructions'
- '+ New Entry' button (primary)
- Table with the fields: Proponent Name (text field), From (MM-DD-YYYY ghost text), To ('Today' for the last entry) and Actions (edit button by default)
THEN all other fields on the modal window are greyed out

### Screenshot for Scenario 2:
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBa1hkIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--760833c91094fdfcef2ddd868dfdf23a3303f340/Screen%20Shot%202023-09-26%20at%204.01.20%20PM.png)

### Scenario 3: 

GIVEN The Modal window with Proponent History is open
WHEN I click on 'Cancel' button
THEN The modal windown closes as no changes have been made

maria-kuzmenko:
For new entries or editing entries - same logic as for #997 and #998

maria-kuzmenko:
prototype

https://www.figma.com/file/j5onJ5RZGAxahxhYp68N8m/Workplan?type=design&node-id=3614-135069&mode=design

Connor1892:
[...]

PR #1626, opened 2024-01-09
#1623 - Implement Create/Update/Delete on Import endpoint of Project

salabh-aot:
#1623 - Implement Create/Update/Delete on Import endpoint of Project
  - Updated import projects API to handle, create, update and delete operations
 
 #999 - Special History - Proponent name
  - fixed issue where relationship holder name was not disabled when editing special field

### DIFF
--- a/epictrack-api/src/api/services/project.py
+++ b/epictrack-api/src/api/services/project.py
@@ -74,7 +74,7 @@ class ProjectService:
             "entity_id": project.id,
             "field_name": "proponent_id",
             "field_value": project.proponent_id,
-            "active_from": project.created_at
+            "active_from": project.created_at,
         }
         SpecialFieldService.create_special_field_entry(proponent_special_field_data)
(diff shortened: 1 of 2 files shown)


epictrack-api/src/api/services/project.py
```diff
@@ -186,14 +189,16 @@ def check_first_nation_available(cls, project_id: int, work_id: int) -> bool:
         return {"first_nation_available": result}
 
     @classmethod
-    def import_projects(cls, file: IO):
+    def import_projects(cls, file: IO):  # pylint: disable=too-many-locals
```
jadmnsaad: this function is too big, can we break it into different small functions with proper naming

salabh-aot: Updated

dinesh-aot (approved):

jadmnsaad (approved):

Issue #1585, opened 2024-01-08
Create Work Modal - Work description required


Connor1892:
**Describe the bug**
Work Description shows as required, but can be saved without it.

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

dinesh-aot:
@djnunez-aot  Please check the back end validation to see if it is required there

PR #1648, opened 2024-01-11
Make Work description mandatory to match backend

djnunez-aot:
#1585 

In the back end work description is mandatory but it is not mandatory in the front end. Updated Create Work form to match.

image

### DIFF
--- a/epictrack-web/src/components/work/WorkForm/index.tsx
+++ b/epictrack-web/src/components/work/WorkForm/index.tsx
@@ -326,7 +326,7 @@ export default function WorkForm({ ...props }) {
           />
         </Grid>
         <Grid item xs={12}>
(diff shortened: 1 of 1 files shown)


epictrack-web/src/components/work/WorkForm/index.tsx
```diff
@@ -35,6 +35,7 @@ const schema = yup.object<Work>().shape({
   federal_involvement_id: yup
     .number()
     .required("Federal Involvement is required"),
+  report_description: yup.string().required("Description is required"),
```
dinesh-aot: can you mention the validation message as "Work description is required"

djnunez-aot: sure thing

dinesh-aot (approved):

Issue #1637, opened 2024-01-10
Remove the Environment Banner from production environment


dinesh-aot:
*Overview*
- The current way is to check the keyword in the window.location.hostname which is not good
- We can have env in the config map(environment variable) - check it for "dev, test, demo or localhost" to show the banner

maria-kuzmenko:
@dinesh-aot , prod still have a test banner

<img width="1535" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOWJJQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--3d4f251a3597cd5f850c222004ebf3b705bad6ab/Screen%20Shot%202024-01-15%20at%209.33.31%20AM.png" alt="Screen Shot 2024-01-15 at 9.33.31 AM.png" />

PR #1652, opened 2024-01-11
Remove the Environment Banner from production environment

djnunez-aot:
#1637 

Remove the banner from the production environment 

-refactor to use env variables instead of window.location

### DIFF
--- a/epictrack-web/src/components/layout/Header/EnvironmentBanner.tsx
+++ b/epictrack-web/src/components/layout/Header/EnvironmentBanner.tsx
@@ -9,8 +9,7 @@ import { AppConfig } from "../../../config";
 const EnvironmentBanner = () => {
   const dispatch = useAppDispatch();
   const env = AppConfig.environment;
-  const isTestEnvironment =
-    env === "dev" || env === "test" || env === "demo" || env === "localhost";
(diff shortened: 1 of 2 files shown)


epictrack-web/src/components/layout/Header/EnvironmentBanner.tsx
```diff
@@ -5,18 +5,17 @@ import { useAppDispatch } from "../../../hooks";
...
 const EnvironmentBanner = () => {
   const dispatch = useAppDispatch();
-  const host = window.location.hostname;
+  const env = AppConfig.environment;
   const isTestEnvironment =
-    host.indexOf("dev") !== -1 ||
-    host.indexOf("test") !== -1 ||
-    host.indexOf("demo") !== -1 ||
-    host.indexOf("localhost") !== -1;
+    env === "dev" || env === "test" || env === "demo" || env === "localhost";
```
dinesh-aot: use includes instead of multiple ===

dinesh-aot: Also, include the env in upper case in the text that we are showing. Like "You are using a TEST environment"

dinesh-aot (approved):

PR #1654, opened 2024-01-11
Add None ministry row

jadmnsaad:
https://app.zenhub.com/workspaces/epictrack-63891ea941d309001fa292cf/issues/gh/bcgov/epic.track/1331

Details:
- Create new migration to add none ministry
- Make responsible ministry required again


saravanpa-aot (commented):
I dont fully understand the context ;but if its for the UI , we should try not to have None data in the database.. but again if this is one of a case where we need such data in DB , we can consider this..would love to discuss this

epictrack-api/migrations/versions/09a79ca8e8fd_add_none_ministry.py
```diff
@@ -0,0 +1,60 @@
...
+    result = conn.execute(
+        sa.text(
+            "SELECT max(id) FROM ministries"
+        )
+    )
+    new_id = result.fetchone()[0] + 1
+
+    op.execute(
+        ministries.insert().values(
+            id=new_id,
```
salabh-aot: We don't need to pass `id` to insert construct. It will fetch the next id value from database  by default.

jadmnsaad:
> I dont fully understand the context ;but if its for the UI , we should try not to have None data in the database.. but again if this is one of a case where we need such data in DB , we can consider this..would love to discuss this

so this is an existing pattern, if you look at the federal_involvements or the substitution_acts tables you will see that same None row. You can see that return from the get codes and displayed in the work form in the federal involvements select and the acts select. I just applied the same logic and pattern

epictrack-api/migrations/versions/09a79ca8e8fd_add_none_ministry.py
jadmnsaad: good point, thanks

saravanpa-aot:
> > I dont fully understand the context ;but if its for the UI , we should try not to have None data in the database.. but again if this is one of a case where we need such data in DB , we can consider this..would love to discuss this
> 
> so this is an existing pattern, if you look at the federal_involvements or the substitution_acts tables you will see that same None row. You can see that return from the get codes and displayed in the work form in the federal involvements select and the acts select. I just applied the same logic and pattern

okay..yeah..i knew i was missing something..thanks for the explanation..

saravanpa-aot (approved):

Issue #1631, opened 2024-01-09
List Management - Project Page - Add Proponent Name


Connor1892:
**As a** user

**I want** to be able to filter projects by proponent

**so that** I can see a full list of projects owned by any given proponent

### Scenario 1: 

GIVEN I am on the Project List
WHEN I start filtering
THEN I see the "Proponent" filter to the right of "Subtype"

dinesh-aot:
- Show the proponent field in the grid

djnunez-aot:
@dinesh-aot does this task include fetching proponents from the backend

dinesh-aot:
> @dinesh-aot does this task include fetching proponents from the backend

I don't think. The proponent field is already there in the project response. Shall we use that
Plus, you can place it between SubType and ENV Region fields in the listing

maria-kuzmenko:
@dinesh-aot weird filter behaviour here

<img width="1185" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMFRkQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--5049f24898e8a59e188fb8871262bca83bdf8cb6/Screen%20Shot%202024-01-18%20at%202.59.16%20PM.png" alt="Screen Shot 2024-01-18 at 2.59.16 PM.png" />

djnunez-aot:
<img width="1128" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMGpoQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--e6e107c7e86e8303c914094b0aaed54cbfdf7385/image.png" alt="image.png" />
Seems like the issue is some of the proponent names are very long should we set a max width ? @dinesh-aot

dinesh-aot:
@djnunez-aot Lets fix the maxWidth to 328 Px.

PR #1664, opened 2024-01-15
Create add proponent field

djnunez-aot:
#1631 

-Add proponent field 

-Fetch proponents by ids

### DIFF
--- a/epictrack-web/src/components/project/ProjectList.tsx
+++ b/epictrack-web/src/components/project/ProjectList.tsx
@@ -15,15 +15,16 @@ import { Restricted, hasPermission } from "../shared/restricted";
 import { ROLES } from "../../constants/application-constant";
 import { searchFilter } from "../shared/MasterTrackTable/filters";
 import { useAppSelector } from "../../hooks";
+import proponentService from "services/proponentService/proponentService";
+import { Proponent } from "models/proponent";
 
 const ProjectList = () => {
   const [envRegions, setEnvRegions] = React.useState<string[]>([]);
   const [subTypes, setSubTypes] = React.useState<string[]>([]);
-  const [proponents, setProponents] = React.useState<string[]>([]);
+  const [proponents, setProponents] = React.useState<Proponent[]>([]);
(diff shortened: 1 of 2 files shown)


epictrack-web/src/components/project/ProjectList.tsx
```diff
@@ -50,6 +52,21 @@ const ProjectList = () => {
...
   const projects = React.useMemo(() => ctx.data as Project[], [ctx.data]);
 
+  async function fetchProponents() {
+    const project_proponents = projects
+      .map((p) => p.proponent_id)
+      .filter((ele, index, arr) => arr.findIndex((t) => t === ele) === index);
+
+    const proponentsResponses = await Promise.all(
+      project_proponents.map((proponentId) =>
+        proponentService.getById(proponentId.toString())
```
jadmnsaad: can we try to find another solution for this other than doing multiple API Calls, one for each proponent Id, can we fetch the whole list from the backend?

epictrack-web/src/components/project/ProjectList.tsx
```diff
@@ -50,6 +52,21 @@ const ProjectList = () => {
 
   const projects = React.useMemo(() => ctx.data as Project[], [ctx.data]);
 
+  async function fetchProponents() {
+    const project_proponents = projects
+      .map((p) => p.proponent_id)
```
jadmnsaad: let's try using full words for variables please for any new implementation

djnunez-aot: sounds good

epictrack-web/src/components/project/ProjectList.tsx
djnunez-aot: yea ill do that then ill create a backend route to get all the ids

dinesh-aot:
@djnunez-aot  We already have proponent info coming from project api response. Make use of that like rest of the column filter and place it between SubType and ENV Region fields. You don't need any back end calls in this ticket. Thanks

dinesh-aot (approved):

PR #1684, opened 2024-01-17
Fixed the issue where team member names where repeated, breaking UI

salabh-aot:
Fixed the issue where team member names where repeated, breaking UI
 - Possible fix for #1536 - RF download bug

- Fixed issue where work started in between report date and end date where causing 500 internal server error


epictrack-api/src/api/reports/resource_forecast_report.py
```diff
@@ -710,3 +614,95 @@ def _add_months(
         if set_to_last:
             result = result.replace(day=monthrange(start.year + year_offset, month)[1])
         return result
+
+    def _get_valid_work_ids(self, report_date: datetime) -> Set[int]:
```
jadmnsaad, line 618: so big, can we break it to more functions please

epictrack-api/src/api/reports/resource_forecast_report.py
```diff
@@ -710,3 +614,95 @@ def _add_months(
...
+                .order_by(WorkPhase.work_id, WorkPhase.phase_id.desc())
+                .distinct(WorkPhase.work_id)
+            )
+            .scalars()
+            .all()
+        )
+        valid_work_ids = set(works_not_finished) & set(works_started)
+        return valid_work_ids
+
+    def _set_month_labels(self, report_date: datetime) -> None:
```
jadmnsaad, line 679: can we break this to more function please

epictrack-api/src/api/reports/resource_forecast_report.py
```diff
@@ -64,9 +51,9 @@ def __init__(self, filters):
             "work_lead",
             "env_region",
             "nrs_region",
-            "staff_first_name",
-            "staff_last_name",
-            "role_id",
+            # "staff_first_name",
+            # "staff_last_name",
```
jadmnsaad, line 55: can we remove these all together instead of leaving them commented out?

Issue #1356, opened 2023-11-27
Delegation Decision Logic


dinesh-aot:
**Overview**
Delegation Decision Logic.
For several WORK, there is a formal “Decision Delegation Event” on who will be responsible for the final WORK DECISION.
When a choice is made in the Decision Delegation Event, the appropriate “Name” needs to be updated to a few locations in the WORK and EVENTs
The “Decision Maker” field in Edit/View WORK modal must be updated to the appropriate “Name” when the ACTION is triggered (the “Name” might come from “CURRENT ACTIVE POSITION ###”)

The “Decision Maker” dropdown in the appropriate (usually final) WORK DECISION (usually the PHASE END EVENT of the LAST PHASE) EVENT modal must be updated to the appropriate “Name” when the ACTION is triggered (the “Name” might come from “CURRENT ACTIVE POSITION ###”)

dinesh-aot:
The chosen decision make should be reflected only on the final decision

Connor1892:
The Final Decision dropdown has the Minister, CEAO, ADM, & EPDs even after assigning it to be the CEAO in a delegation milestone

PR #1700, opened 2024-01-18
Last phase last decision event should show selected decision maker

dinesh-aot:
#1356 
- works/:work_id/phases - should return additional field called is_last_phase

### DIFF
--- a/epictrack-api/src/api/services/work_phase.py
+++ b/epictrack-api/src/api/services/work_phase.py
@@ -109,15 +109,12 @@ class WorkPhaseService:  # pylint: disable=too-few-public-methods
         result_dict = {}
         work_ids = list(work_params_dict.keys())
 
-        work_phases_dict, total_work_phases = cls._query_work_phases(work_ids)
+        work_phases_dict = cls._query_work_phases(work_ids)[0]
 
-        index = 1
         for work_id, work_phase_id in work_params_dict.items():
             result_dict[work_id] = cls._find_work_phase_status(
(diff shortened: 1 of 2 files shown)


epictrack-api/src/api/services/work_phase.py
```diff
@@ -109,14 +109,14 @@ def find_multiple_works_phases_status(
...
+        work_phases_dict = cls._query_work_phases(work_ids)[0]
 
         index = 1
         for work_id, work_phase_id in work_params_dict.items():
             result_dict[work_id] = cls._find_work_phase_status(
                 work_id, work_phase_id, work_phases_dict.get(work_id, [])
             )
-            result_dict["is_last_phase"] = index == total_work_phases
+            # result_dict["is_last_phase"] = index == total_work_phases
             index = index + 1
```
salabh-aot: I think we can remove `index = index + 1` from here since it has no use inside the loop anymore.

epictrack-api/src/api/services/work_phase.py
```diff
@@ -150,6 +150,7 @@ def _find_work_phase_status(cls, work_id, work_phase_id, work_phases):
         events = EventService.find_events(work_id, event_categories=PRIMARY_CATEGORIES)
         if work_phase_id is not None:
             work_phases = [wp for wp in work_phases if wp.id == work_phase_id]
+        index  = 1
         for work_phase in work_phases:
```
salabh-aot: This could be replaced with 
```
for index, work_phase in enumerate(work_phases, start=1):
    ...
```

And avoid manually initializing and updating index.

Issue #1687, opened 2024-01-17
Anticipated Report - Updated Logic


salabh-aot:
**Anticipated Report Logic**

IF WORK ACTIVE = TRUE								AND
IF WORK STATE = In Progress  OR Suspended				AND
IF any (EventType = 5) REFERRAL MILESTONE DATE(anti/actu) equal to or greater than [REPORT DATE - 7]	THEN
Display on Anticipated EA Schedule

Connor1892:
this needs an estimate

dinesh-aot:
**Test Failed in the below scenario**
- Go to all works and check the below mentioned work
  - Debolt Saline Water - Minister's Designation - 111
  - Go to the last phase and check the decision milestone
  - The Referral Decision milestone is anticipated on Feb25 2024
- Go to Referral Report and choose Jan 8 2024
- See if you are able to see the above mentioned project in the report
- 

Pre-Open the accordions in the initial load

salabh-aot:
The mentioned work don't have any status updates. As per current conditions, we are looking at mandatory (`JOIN`) status update with `approved = True and posted_date >= report_start_date`. Is the condition correct? Or status updates are not mandatory?

Note: With non-mandatory status updates, the `date_updated & additional_info` could be `null` or empty.

dinesh-aot:
@salabh-aot 
The status query can be outerjoin instead of inner join. 
Status query shouldn't be a criteria to select the work in the report, instead, it should be used for showing the staleness.
In frontend, the staleness should be according to the below mentioned logic
(Today - LastStatusUpdateDate).Days > 10 days = RED
(Today - LastStatusUpdateDate).Days  is between10 - 6  = Yellow
(Today - LastStatusUpdateDate).Days is less than equal to 5 = GREEN

In case if there is no status update at all, In front end, instead of the date field to show colors, it should say "Needs Status" in RED color

PR #1702, opened 2024-01-18
Anticipated Report - Updated Logic #1687

salabh-aot:
Anticipated report refactored according to following conditions
  - WORK ACTIVE = TRUE
  - WORK STATE = In Progress OR Suspended
  - any (EventType = 5) REFERRAL MILESTONE DATE(anti/actu) equal to or greater than [REPORT DATE - 7]


epictrack-api/src/api/reports/anticipated_schedule_report.py
```diff
@@ -228,3 +164,95 @@ def generate_report(self, report_date, return_type):
...
+        return (
+            db.session.query(
+                Event.work_id,
+                func.min(Event.anticipated_date).label("min_anticipated_date"),
+            )
+            .join(
+                EventConfiguration,
+                and_(
+                    Event.event_configuration_id == EventConfiguration.id,
+                    EventConfiguration.event_type_id == 5,
```
dinesh-aot: Instead of hard coding this number 5, I think we have an enum in the model folder

epictrack-api/src/api/reports/anticipated_schedule_report.py
```diff
@@ -150,34 +93,27 @@ def _fetch_data(self, report_date):
             .join(Region, Region.id == Project.region_id_env)
             .join(EAAct, EAAct.id == Work.ea_act_id)
             .join(Ministry)
+            .join(latest_status_updates, latest_status_updates.c.work_id == Work.id)
```
dinesh-aot, line 96: Have we discussed joining it with status update query. Cz I think as per the latest logic, the result has nothing to do with the status update

PR #1705, opened 2024-01-18
Add ProjectList Unit Tests

djnunez-aot:
#1618 

- Create unit test for ProjectList component

### DIFF
--- a/epictrack-web/src/components/project/__test__/ProjectList.cy.tsx
+++ b/epictrack-web/src/components/project/__test__/ProjectList.cy.tsx
@@ -1,0 +1,99 @@
+import { MemoryRouter as Router } from "react-router-dom";
+import ProjectList from "../ProjectList";
+import { Type } from "models/type";
+import { SubType } from "models/subtype";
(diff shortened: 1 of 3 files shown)


epictrack-web/src/components/project/__test__/ProjectList.cy.tsx
```diff
@@ -0,0 +1,126 @@
...
+        <MasterContext.Provider value={createMockContext()}>
+          <ProjectList />
+        </MasterContext.Provider>
+      </Router>
+    );
+  });
+
+  it("should display the project list", () => {
+    // Select the table container
+    cy.get(".MuiTableContainer-root.css-3i8oh7-MuiTableContainer-root").should(
```
jadmnsaad: This is harder to maintain. Let's just add data-cy to those component and get by it. Let's not get by the MUI class names

jadmnsaad (changes requested):
get by data-cy instead of MUI classnames

jadmnsaad (approved):

Issue #1640, opened 2024-01-10
Work "Title" is 80 Characters - Should it be 80 or more than that. Lets set it to 150


dinesh-aot:
- Change the front end validation
- Put the character counter above the text field
- Change the character limit in the back end using migration

dinesh-aot:
- Change the front end validation
- Put the character counter above the text field
- Change the character limit in the back end using migration

djnunez-aot:
@dinesh-aot In the back end title is already set to 150 
<img width="818" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBL3pCQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--8765f8db5ba790584fa0898f5e8a9b134e1cf9f3/image.png" alt="image.png" />

created a pr https://github.com/bcgov/EPIC.track/pull/1659 to address the first two bullet points

maria-kuzmenko:
@dinesh-aot work Title goes into negatives now. Did we discuss to stop at 150?

<img width="1497" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBenJkQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--47aa9316fcea0c19f36d534d89ba4a68fbbc77b0/negativecount.png" alt="negativecount.png" />

PR #1712, opened 2024-01-19
Add max length to ControlledTextField

djnunez-aot:
#1640 

-Add max length to prevent user from typing more if the textfield is at max value


epictrack-web/src/components/shared/controlledInputComponents/ControlledTextField.tsx
```diff
@@ -31,6 +33,9 @@ const ControlledTextField: FC<IFormInputProps> = ({
           {...field}
           {...otherProps}
           onChange={(e) => {
+            if (maxLength && e.target.value.length > maxLength) {
+              return;
```
dinesh-aot: I think you don't want to check the max length like this. You should be able to set it like this inside the common control
inputProps={{
            maxLength: 12,
          }}

dinesh-aot (approved):

PR #1718, opened 2024-01-22
Restrict Status and Issues actions by role and team membership

jadmnsaad:
https://app.zenhub.com/workspaces/epictrack-63891ea941d309001fa292cf/issues/gh/bcgov/epic.track/1586

Details:
- wrap buttons with Restricted component in UI
- check for role and membership in service methods in backend


jadmnsaad:
@dinesh-aot as per my conversation with @saravanpa-aot, we need a new ticket to address the failure in tests

epictrack-api/src/api/models/work_issues.py
```diff
@@ -41,7 +41,7 @@ class WorkIssues(BaseModelVersioned):
         "WorkIssueUpdates",
         back_populates="work_issue",
         lazy="joined",
-        order_by="desc(WorkIssueUpdates.posted_date)",
+        order_by="desc(WorkIssueUpdates.created_at)",
```
salabh-aot, line 44: Is this discussed with Geoff? I think the requirement was, issue updates could be made with past dates (hence `posted_date`). Sorting by `created_at` could lead to them being not ordered as expected.

jadmnsaad: If we should use posted_date to order we should restrict users input to be minimum 1 day + last most recent date  to prevent reshuffling of ordering. We do not have that now so having them order by posted_date at this point adds a bug. When I added the posted_date field I changed this ordering to posted_date from created_date. I am reverting this change now because of the bug.

I discussed with Maria the minimum date logic, I should need to add that first then revert this back to order by posted_date

PR #1730, opened 2024-01-23
made header as a fixture and removed mocking of auth

saravanpa-aot:
https://github.com/bcgov/EPIC.track/issues/1720

Took this approach since this is simpler and mocks closes to an actual jwt call.
If the approach is okay , i will extend it to all the tests

### DIFF
--- a/epictrack-api/tests/conftest.py
+++ b/epictrack-api/tests/conftest.py
@@ -159,7 +159,7 @@ def session(app, db):
 
 
 @pytest.fixture()
-def staff_admin_headers(client, jwt, request):
+def auth_header(client, jwt, request):
     """Create a basic admin header for tests."""
     staff_user = TestJwtClaims.staff_admin_role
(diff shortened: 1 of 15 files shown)


epictrack-api/tests/conftest.py
```diff
@@ -159,14 +158,9 @@ def restart_savepoint(sess2, trans):
...
-    @wraps(f)
-    def decorated_function(*args, **kwargs):
-        return f(*args, **kwargs)
-
-    return decorated_function
-
-
-patch("api.utils.auth.require", mock_decorator, spec=True).start()
+@pytest.fixture()
+def staff_admin_headers(client, jwt, request):
```
dinesh-aot: Shall we have a fixture factory created else where and call it in the conftest instead of this. Better we maintain multiple classes for individual headers and stack it in single class and call the configure() method of that class in conftest. In that way we don't want to over crowd the conftest with such specific impelementations

dinesh-aot (approved):

PR #1733, opened 2024-01-24
Add work simple title

jadmnsaad:
https://app.zenhub.com/workspaces/epictrack-63891ea941d309001fa292cf/issues/gh/bcgov/epic.track/1630

Details:
- Add simple_title in Work
- Add migration with script to populate simple title from full title
- Make project title and worktype mandatory and display as unchangeable text in the title input
- Update resource schemas to have simple_title


epictrack-api/tests/utilities/factory_scenarios.py
```diff
@@ -31,7 +31,7 @@ class TestProjectInfo(Enum):
         "name": fake.word(),
         "description": fake.sentence(),
         "address": fake.address(),
-        "type_id": fake.random_int(min=1, max=10),
+        "type_id": fake.random_int(min=1, max=9),
```
dinesh-aot, line 34: any particular reason for this

jadmnsaad: because of unit test failure.

there is no 10th type, when type 10 was picked Foreign key constraint error was being thrown. type_id = 10 is not valid it doesn't exist

epictrack-api/migrations/versions/74e58a15d180_add_simple_title_to_work.py
```diff
@@ -0,0 +1,48 @@
...
+revision = '74e58a15d180'
+down_revision = '2edbfb8b9c0e'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    with op.batch_alter_table('works', schema=None) as batch_op:
+        batch_op.add_column(sa.Column('simple_title', sa.String(length=30), nullable=True))
```
dinesh-aot, line 22: Just a thought, is it possible to have the "title" a derived field. If you look at the work model, workType and project are relationship there. Is it possible to derive it when required rather than saving it redundantly. @saravanpa-aot @salabh-aot thoughts?

jadmsaadaot: would you prefer that the current solution be blocked until we have this discussion?

salabh-aot: That is possible. We have implemented something similar for Staff's full name.

jadmnsaad: how are we moving forward?

1 - Merge this and do the derived field logic in a separate step (easier for me)
2 - Block this merge and do it here

dinesh-aot (approved):

Issue #1732, opened 2024-01-23
Project Modal - Abbreviation Field


Connor1892:
**Describe the bug**
When Brandon Wulff (a superuser) tries to fill in the Abbreviation Field, he gets 1 character in then the field disables.

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

PR #1734, opened 2024-01-24
Abbreviation input was taking user input even though user doesn't have permission

dinesh-aot:
#1732 
User without "extended_edit" permission tries to change abbreviation field in project form can type first alphabet then the textbox gets disabled.
This was due to the fact that "exception" property was set in the <Restricted component which was causing a false positive condition triggered inside on the initial load

### DIFF
--- a/epictrack-web/src/components/project/ProjectForm/index.tsx
+++ b/epictrack-web/src/components/project/ProjectForm/index.tsx
@@ -458,7 +458,6 @@ export default function ProjectForm({ ...props }) {
             <Restricted
               allowed={[ROLES.EXTENDED_EDIT]}
               errorProps={{ disabled: true }}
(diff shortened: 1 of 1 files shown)


jadmnsaad:
We need to consider the requirement here https://app.zenhub.com/workspaces/epictrack-63891ea941d309001fa292cf/issues/gh/bcgov/epic.track/1201

image

so it if wasn't made and the system failed to generate one, users with Edit should be able to add one. Without the exception now it is always restricted to non-superusers

jadmnsaad (commented):
We need to allow users with Edit to add to that field if no abbreviation was generated. Or check with  business if this is not a requirement

jadmnsaad (approved):

PR #1735, opened 2024-01-24
Add max width filtering for filters

djnunez-aot:
#1631 
Add max width property to filter select to deal with elements being too long for the filter list

### DIFF
--- a/epictrack-web/src/components/shared/filterSelect/type.ts
+++ b/epictrack-web/src/components/shared/filterSelect/type.ts
@@ -24,6 +24,7 @@ declare module "react-select/dist/declarations/src/Select" {
       getOptionLabel?: (option: any) => string;
       getOptionValue?: (option: any) => string;
       label?: string;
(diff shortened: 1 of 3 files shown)


epictrack-web/src/components/shared/filterSelect/type.ts
```diff
@@ -24,6 +24,7 @@ declare module "react-select/dist/declarations/src/Select" {
       getOptionLabel?: (option: any) => string;
       getOptionValue?: (option: any) => string;
       label?: string;
+      maxWidth?: string;
```
jadmnsaad, line 27: Do we have to define and pass these props now explicitly? Is the example in the ControlledTextField too hard to achieve here?

image

djnunez-aot: how do we pass that into the style prop of the select field its the maxwidth of the component not the character length

dinesh-aot (approved):

Issue #955, opened 2023-09-15
All works - updating work modal - federal involvement


TomChapmanGov:
---
If you create a work with no federal involvement then try to update it, it will give an error unless you select an option for federal act.

When creating a work, if federal involvement is set to none then the federal act should also be defaulted to none.
Make the federal act a required field

---

**As a** superuser

**I want** federal involvement should be tied to federal act

**so that** when Federal Involvement is set to none, Federal Act is set to none

### Scenario 1: 

GIVEN I am on the create work modal
WHEN Federal Involvement is none
THEN Federal Act is none

### Scenario 2:
 xxxxxx

### Validation Rules? 

(If yes, list here)

### Screenshot 

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBdSttIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--bc073aedc30d794f60fe625f92c6a2a2ccdcc2b3/image.png)

(Link above will always be the most up-to-date version.)

PR #1748, opened 2024-01-25
Make Federal Act default to None when Federal Involvement is set to None

djnunez-aot:
#955 
-Federal act is set to none when federal involvement is set to none

-Add correct text when federal act field is empty

### DIFF
--- a/epictrack-web/src/components/work/WorkForm/index.tsx
+++ b/epictrack-web/src/components/work/WorkForm/index.tsx
@@ -138,8 +138,7 @@ export default function WorkForm({ ...props }) {
     reset(ctx.item ?? defaultWork);
   }, [ctx.item]);
 
-  useEffect(() => {
-    // Find the object where name is 'None'
+  React.useEffect(() => {
     const noneFederalInvolvement = federalInvolvements.find(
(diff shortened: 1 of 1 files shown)


epictrack-web/src/components/work/WorkForm/index.tsx
```diff
@@ -133,12 +134,27 @@ export default function WorkForm({ ...props }) {
...
+      ({ name }) => name === "None"
+    );
+
+    if (
+      noneSubstitutionAct &&
+      Number(federal_involvement_id) === noneFederalInvolvement?.id
+    ) {
+      setValue("substitution_act_id", noneSubstitutionAct?.id);
+    }
+  }, [federal_involvement_id, setValue]);
```
dinesh-aot: Do we need this setValue in the dependency list. I am not sure. Could you please make sure it works without this dependency.

djnunez-aot: sure

djnunez-aot: it works just tested it i removed the unused dependency

dinesh-aot (approved):

Issue #1698, opened 2024-01-18
Remove the edit button from Team Tab Team info box from work plan


dinesh-aot:
**Overview**
It is decided to remove the Edit and show corresponding pop up of work lead or epd or team from Team Info pane in Team tab in work plan.

Connor1892:
the mouse still recognizes a button there. the button seems to be invisible but the hotspot still exists.

PR #1753, opened 2024-01-25
team info edit button removal, migration fix, work title display getting undefined at times

dinesh-aot:
#1698 
- Migration fixed. Using query instead of Model class
- Commented the whole edit button box in team tab
- Work title display logic gets undefind at times(if the projects, worktypes not loaded completely)

### DIFF
--- a/epictrack-web/src/components/work/WorkForm/index.tsx
+++ b/epictrack-web/src/components/work/WorkForm/index.tsx
@@ -100,8 +100,8 @@ export default function WorkForm({ ...props }) {
     watch,
   } = methods;
 
-  const work_type_id = watch("work_type_id");
-  const project_id = watch("project_id");
+  const workTypeId = watch("work_type_id");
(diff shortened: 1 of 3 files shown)


epictrack-web/src/components/work/WorkForm/index.tsx
```diff
@@ -224,7 +222,7 @@ export default function WorkForm({ ...props }) {
   useEffect(() => {
     const prefix = getTitlePrefix();
     setTitlePrefix(prefix);
-  }, [work_type_id, project_id]);
+  }, [workTypeId, projectId, projects, workTypes]);
```
jadmnsaad, line 243: right of course because getTitlePrefix is depending on projects and workTypes, good catch. I also suggest wrapping the inside of the useEffect by if(projects && workTypes) so we don't do multiple state updates, we just do once, once they are loaded

jadmnsaad (approved):

PR #1759, opened 2024-01-26
Add Consultation Level and remove PIN

jadmnsaad:
https://app.zenhub.com/workspaces/epictrack-63891ea941d309001fa292cf/issues/gh/bcgov/epic.track/1722

Details:
- Add model, service, and resource for indigenous nations consultation levels
- Add migration scripts, add logic to populate consultation level in work by PIN choice
- Remove PIN column and logic
- Add the consultation level select to the form and to the table


epictrack-api/src/api/models/indigenous_consultation_levels.py
```diff
@@ -0,0 +1,45 @@
...
+        """Return Json representation."""
+        return {
+            'id': self.id,
+            'name': self.name,
+            'sort_order': self.sort_order,
+            'is_active': self.is_active
+        }
+
+    @classmethod
+    def find_all_active_consultation_levels(cls):
```
dinesh-aot: Is there a method already there in the base class for getting all active and non-deleted items?

jadmnsaad: removed

PR #1760, opened 2024-01-29
Resource forecast refactored

dinesh-aot:


epictrack-api/src/api/reports/resource_forecast_report.py
```diff
@@ -183,15 +184,146 @@ def _get_report_meta_data(self, report_date: datetime, available_width: float):
...
+            Project.query.filter(
+                Project.is_project_closed.is_(False),
+                Project.is_deleted.is_(False),
+                Project.is_active.is_(True),
+            )
+            .join(
+                Work,
+                and_(
+                    Work.project_id == Project.id,
+                    Work.work_state.in_(
```
salabh-aot, line 205: We can remove this duplicate `Work.work_state.in_(` condition since it is given in `filter()` below.

PR #1762, opened 2024-01-29
Rich text editor unit tests

djnunez-aot:
#1603 

-Add unit tests for rich text editor

### DIFF
--- a/epictrack-web/src/components/shared/richTextEditor/__test__/RichTextEditor.cy.tsx
+++ b/epictrack-web/src/components/shared/richTextEditor/__test__/RichTextEditor.cy.tsx
@@ -1,0 +1,26 @@
+import RichTextEditor from "../index";
+
+describe("RichTextEditor Component Tests", () => {
+  it("should render the component", () => {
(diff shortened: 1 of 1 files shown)


jadmnsaad:
Boss @djnunez-aot , on the topic of RichTextEditor, can you see if you can create a ControlledRichTextEditor. Check stuff like ControlledTextField, or ControlledDatePicker.

@dinesh-aot I think this is valuable because we do use the rich text editor in some forms but kinda in a detached way which isn't great

jadmnsaad (approved):

dinesh-aot:
> Boss @djnunez-aot , on the topic of RichTextEditor, can you see if you can create a ControlledRichTextEditor. Check stuff like ControlledTextField, or ControlledDatePicker.
> 
> @dinesh-aot I think this is valuable because we do use the rich text editor in some forms but kinda in a detached way which isn't great

Yeah, I will create a new card for that and let david work on it.

dinesh-aot (approved):

PR #1767, opened 2024-01-29
Fix design and consistency issues in Status and Issues tabs

jadmnsaad:
https://app.zenhub.com/workspaces/epictrack-63891ea941d309001fa292cf/issues/gh/bcgov/epic.track/1752

Details:
- apply design tweaks and fixes in status tabs
- Apply design tweaks and fixes in issues tab

Purely UI work

### DIFF
--- a/epictrack-web/src/components/shared/index.tsx
+++ b/epictrack-web/src/components/shared/index.tsx
@@ -426,15 +426,26 @@ export const GrayBox = ({ children, sx, ...rest }: GrayBoxProps) => {
   );
 };
 
-export const ETPreviewText = styled(Typography)((props) => ({
-  fontSize: "14px",
-  fontStyle: "normal",
-  fontWeight:
-    props.fontWeight === "bold"
-      ? MET_Header_Font_Weight_Bold
-      : MET_Header_Font_Weight_Regular,
-  lineHeight: "21px",
-}));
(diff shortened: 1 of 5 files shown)


epictrack-web/src/components/shared/index.tsx
```diff
@@ -426,10 +426,13 @@ export const GrayBox = ({ children, sx, ...rest }: GrayBoxProps) => {
   );
 };
 
-export const ETPreviewText = styled(Typography)(() => ({
+export const ETPreviewText = styled(Typography)((props) => ({
   fontSize: "14px",
   fontStyle: "normal",
-  fontWeight: "400",
+  fontWeight:
+    props.fontWeight === "bold"
```
dinesh-aot: If it is either bold or regular, then, can we consider a boolean parameter instead of fontWeight. Something like isBold.

jadmnsaad: I actually chose to do it this way because to accept a custom isBold parameter this component will have to be changed from a styled component and we have to implement it differently, I thought this would be easier. fontWeight is a default style parameter and I just overrode it

jadmnsaad: anyway, changed the implementation to like the other custom topographies and passed bold as a prop

dinesh-aot (approved):

PR #1782, opened 2024-01-31
Resource forecast related changes

dinesh-aot:
- clean ups
- sort logic fix
- front end table pagination


epictrack-api/src/api/reports/resource_forecast_report.py
```diff
@@ -475,7 +475,7 @@ def _sort_data(self, data, second_phases) -> List:
         exemption_orders = self._sort_data_by_work_type(
             data, WorkTypeEnum.EXEMPTION_ORDER.value
         )
-        amendments = self._sort_data_by_work_type(data, WorkTypeEnum.AMENDMENT.value)
+        amendments = self._sort_data_by_work_type(data, WorkTypeEnum.AMENDMENT.value, second_phases)
```
salabh-aot, line 478: Sorry, I forgot this part, we need to add this to all calls to `_sort_data_by_work_type`

Issue #1787, opened 2024-01-31
MyWorkPlan - Non Staff View


dinesh-aot:
**Overview**
When a non-staff user login to the application
Go to MyWorkPlan Page
Since the user is not a staff, there should be no work plans listed for him if he chooses to turn the toggle on

Connor1892:
@dinesh-aot what am I testing here?

dinesh-aot:
@Connor1892  I just updated the description above. Let me know if you need help in testing this.

PR #1788, opened 2024-02-01
myworkplan when non-staff logged in

dinesh-aot:
#1787 
- Show nothing in myworkplan when non-staff member logged in
- Fixed filter select returning <SELECT_ALL> as selected value

### DIFF
--- a/epictrack-api/src/api/models/work.py
+++ b/epictrack-api/src/api/models/work.py
@@ -158,7 +158,7 @@ class Work(BaseModelVersioned):
 
     @classmethod
     def _filter_by_staff_id(cls, query, staff_id):
-        if staff_id:
+        if staff_id is not None:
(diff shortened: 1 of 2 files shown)


saravanpa-aot (approved):

epictrack-api/src/api/models/work.py
```diff
@@ -158,7 +158,7 @@ def filter_by_search_criteria(cls, query, search_filters: WorkplanDashboardSearc
 
     @classmethod
     def _filter_by_staff_id(cls, query, staff_id):
-        if staff_id:
+        if staff_id is not None:
```
jadmnsaad, line 161: We tried this solution first together and I remember it had an issue, is that resolved?

jadmnsaad: I tried it again and it's working with this solution. It seems this is just fine.

jadmnsaad (approved):

PR #1796, opened 2024-02-02
Filter select unit tests

djnunez-aot:
#1596 

-unit tests for the filter select component


jadmnsaad (approved):

epictrack-web/src/components/shared/__test__/FilterSelect.cy.tsx
```diff
@@ -0,0 +1,40 @@
...
+      { label: "Option 2", value: "2" },
+    ];
+
+    cy.mount(<FilterSelect options={options} variant={"inline"} />);
+
+    cy.get("input").should("exist");
+  });
+
+  it("can select an option", () => {
+    const options = [
```
dinesh-aot, line 17: You can also include tests for FilterSelect by setting isMulti option true. In that you can test the "ClearSelect, Cancel, Apply" ...etc

Issue #1792, opened 2024-02-01
Resource Forecast report download failed


dinesh-aot:

maria-kuzmenko:
@dinesh-aot , resource forecast doesn't retrieve anything - I clicked on 2024-02-08, got an error message and the spinning wheel is forever there.

dinesh-aot:
@maria-kuzmenko @EAO-App-Admin 
HVC 2040 - EAC Assessment in test environment has second phase inactive due to the decision taken in the pre-ea phase.
We will not have a second phase moving forward. Early Engagement is going to be the first phase. the sort logic will have to be changed accordingly

PR #1829, opened 2024-02-09
EventConfigurations fix for assessment 

dinesh-aot:
**Overview**
Decision phase didn't have START and END events configured which caused the resource forecast report failure
#1792


epictrack-api/migrations/versions/735a693b4a59_fix_to_existing_event_configurations.py
```diff
@@ -0,0 +1,31 @@
...
+revision = '735a693b4a59'
+down_revision = '5b390f888c3f'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.execute("update event_configurations ec set event_position = 'START' where ec.template_id in (select id from event_templates et where et.phase_id = 20 and name = 'Referral Package Received by Ministers' order by sort_order)")
+    op.execute("update event_configurations ec set event_position = 'END' where ec.template_id in (select id from event_templates et where et.phase_id = 20 and name = 'EAC Ministers Decision' order by sort_order)")
```
salabh-aot: Could Phase ID differ depending on the order we import event templates?

salabh-aot (changes requested):

epictrack-api/migrations/versions/735a693b4a59_fix_to_existing_event_configurations.py
dinesh-aot: That is a possibility. I will put another nested query for this.

Issue #752, opened 2023-06-22
Celebratory Fireworks when Phase completed


Connor1892:
---

when phase closes the accordion should move to closed phases with toggle flipped to showing.

celebration animation shows every time the work is opened for 24hours after the phase is closed

https://alampros.github.io/react-confetti/?path=/story/props-demos--knobs

---

**As a** (type of user)

**I want** (goal)

**so that** (reason/ why/ value)

### Scenario 1: 

GIVEN ... (you or your condition)
WHEN ... (what you do)
THEN ... (what you see)

### Scenario 2:
 xxxxxx

### Validation Rules? 

(If yes, list here)

### Screenshot 

(Link above will always be the most up-to-date version.)

Connor1892:
confetti for phase?
Fireworks for work?

dinesh-aot:
@djnunez-aot Can you take a look at the below mentioned library for the fireworks
https://www.kirilv.com/canvas-confetti/
It has wide verities of fireworks and animations. 
Regarding the implementation, these are my thoughts @jadmsaadaot  Please add your comments
- Create both fireworks and confetty as components or single components with passing parameters to distinguish
- Fireworks should be shown in the entire application level, so can it be in the root level and triggered by redux state and keep it for a specific duration and goes off
- Show confetty only when a phase gets completed, it can be shown within the phase accordion

Connor1892:
@djnunez-aot MARIA IS SO BETRAYED BY THE LACK OF FIREWORKS AND CONFETTI!!!!

Connor1892:
no displays for both for old works and newly created.

PR #1835, opened 2024-02-12
Add fireworks for phase completion

djnunez-aot:
#752 

-Add fireworks when celebratory phase is completed


epictrack-web/src/components/workPlan/event/EventList.tsx
```diff
@@ -324,11 +318,25 @@ const EventList = () => {
...
     setTaskEvent(undefined);
     setMilestoneEvent(undefined);
+    getTemplateUploadStatus()
+      .then(() => {
+        return getWorkById();
+      })
+      .then(() => {
+        if (
+          milestoneEvent &&
+          milestoneEvent.event_configuration.event_position ===
```
dinesh-aot, line 331: You should add one more condition to make sure user has actually completed the phase by checking milestoneEvent.actual_date.

Issue #1681, opened 2024-01-16
Refactor services/event.py to reduce lines of code


jadmnsaad:
pylint is now complaining about it <img width="810" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeWJRQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--85b8675f993eafd9f32c585485fa050f1e55e8e8/image.png" alt="image.png" />

PR #1843, opened 2024-02-13
Event Service Refinements and SKIP_EVENT_LOGIC env added

dinesh-aot:
#1681 
- Refactored event.py
- Added a few comments
- Added SKIP_EVENT_LOGIC env to skip event logic for the purpose of historical data entry


epictrack-api/src/api/services/event.py
```diff
@@ -45,15 +46,14 @@
 from api.utils import util
 from api.application_constants import MIN_WORK_START_DATE
 
+from api.config import _Config
```
salabh-aot: We can get the current config for the environment by
```
from flask import current_app
config = current_app.config
```
This way can avoid importing a 'private' member.

salabh-aot (changes requested):

salabh-aot (approved):

Issue #1309, opened 2023-11-22
Avoid handling certain entities as "Code" - (Standard api that provide most of the master data as name/value pair)


dinesh-aot:
**Overview**
The current list of models handling as "Code"
 - [x] "positions"
 - [x]  "projects"
 - [x]  "ministries"
 - [x]  "proponents"
 - [x] "ea_acts"
 - [x] "work_types"
 - [x]  "eao_teams"
 - [x]  "regions"
 - [x]  "types"
 - [x]  "phases"
 - [x]  "sub_types"
 - [x] "federal_involvements"
 - [x]  "responsibilities"
 - [x]  "roles"
  - [x]  "substitution_acts"
 - [x]  "pip_org_types";
 
We have avoid using "Code" for the below ones
- projects
- proponents

PR #1847, opened 2024-02-15
Avoid handling certain entities as "Code"

salabh-aot:
#1309
  - Switched following resources from codes API
        - positions
    	- projects 
    	- ministries 
    	- proponents 
    	- ea_acts 
    	- work_types 
    	- eao_teams 
    	- regions 
    	- types 
    	- phases 
    	- sub_types 
    	- federal_involvements 
    	- responsibilities 
    	- roles 
    	- substitution_acts 
    	- pip_org_types;


saravanpa-aot (commented):
I like the part that we have separate API;s ;but at the same time i am just worried about our code duplication

dinesh-aot:
> I like the part that we have separate API;s ;but at the same time i am just worried about our code duplication

What you said is correct and for this reason plus making it available for usage in short period of time we went with Code apis initially. A few advantage of separating it
- Clearer understanding of each of those endpoints
- Will be able to apply caching or logging separately to individual endpoints considering their parameters
- Expecting to build a UI around these entities in future(We can think about flask admin for sure.). This will be useful at that point of time
- We are already using it everywhere as Code apis, it would be difficult to change this in future. Plus, this wasn't a good idea to create it as single API initially. We went that way to build the prototype using formsflow

saravanpa-aot:
> > I like the part that we have separate API;s ;but at the same time i am just worried about our code duplication
> 
> What you said is correct and for this reason plus making it available for usage in short period of time we went with Code apis initially. A few advantage of separating it
> 
> * Clearer understanding of each of those endpoints
> * Will be able to apply caching or logging separately to individual endpoints considering their parameters
> * Expecting to build a UI around these entities in future(We can think about flask admin for sure.). This will be useful at that point of time
> * We are already using it everywhere as Code apis, it would be difficult to change this in future. Plus, this wasn't a good idea to create it as single API initially. We went that way to build the prototype using formsflow

great points.thanks for the exaplanation..lets merge it

dinesh-aot (approved):

Issue #1190, opened 2023-11-08
Workplan Page - days left warning


Connor1892:
---

the days left area of the accordion, when it has a warning, is not very noticeable in its dark yellow.
we need to make it pop more to show that a warning is on display

---

**As a** user
**I want** see instantly what phases are out of time
**so that** I can add extension milestone to extend the phase timeline

### Scenario 1: 

GIVEN My Workplans is open
WHEN I click on any Work which timeline is beyond legislative timeline
THEN I see days left in Error/Dark rgba (163.30,34,1) and a exclamation symbol indicator 

### Screenshot 
<img width="1429" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeXA5QXc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--fa784a44cf6fda298cf7472e9575d88ca9959749/Screen%20Shot%202024-02-21%20at%202.50.51%20PM.png" alt="Screen Shot 2024-02-21 at 2.50.51 PM.png" />

Connor1892:
@maria-kuzmenko colour code for new warning colour pls!

maria-kuzmenko:
@djnunez-aot , link to the design - https://www.figma.com/file/XcYynwH7xrO2P30uqewaLS/For-Development?type=design&node-id=41-142&mode=dev

What should be done:
1. Exclamation indicator next to the status should be recoloured into
    background: rgba(252, 237, 238, 1)
     exclamation mark: rgba(163, 30, 34, 1)

2. This component we are reusing and change a triangle to it

PR #1871, opened 2024-02-22
Change workplan page colors

djnunez-aot:
#1190 

-Change workplan page exclamation icon colors to 

pink:
#fcedee
red:
#a31e22


jadmnsaad (approved):

jadmnsaad:
ideally these colors should be set via a prop, I think there are examples in other icons

djnunez-aot:
> ideally these colors should be set via a prop, I think there are examples in other icons

Okay ill import the palette there i didn't think i saw the palette being used but will do that now before merge

jadmnsaad:
> > ideally these colors should be set via a prop, I think there are examples in other icons
> 
> Okay ill import the palette there i didn't think i saw the palette being used but will do that now before merge

I see you have used the Palette, that's good. But the color should still be passed as a prop when the icons are used not hard coded. And you can do that only for the exclamation icon for this task, you don't need to change it for all

dinesh-aot (approved):

Issue #1175, opened 2023-11-06
Button Enhancements


Connor1892:
---

issues tab report preview buttons
- sizes change when flipping between 30-60-90 and referral schedule previews

---

**As a** (type of user)

**I want** (goal)

**so that** (reason/ why/ value)

### Scenario 1: 

GIVEN ... (you or your condition)
WHEN ... (what you do)
THEN ... (what you see)

### Scenario 2:
 xxxxxx

### Validation Rules? 

(If yes, list here)

### Screenshot 

(Link above will always be the most up-to-date version.)

maria-kuzmenko:
@djnunez-aot 

not working for Issue Tab, should work the same way as on the Status Tab

PR #1875, opened 2024-02-22
Fix issue buttons

djnunez-aot:
#1175 

-Fix inconsistent padding toggling between buttons in workplan issues tab

### DIFF
--- a/epictrack-web/src/components/shared/TabButton.tsx
+++ b/epictrack-web/src/components/shared/TabButton.tsx
@@ -1,0 +1,32 @@
+import React from "react";
+import { Button } from "@mui/material";
+import { Palette } from "styles/theme";
+
+type TabButtonProps = {
+  children: React.ReactNode;
+  tabIndex: number;
+  selectedTabIndex: number;
+  setSelectedTabIndex: (index: number) => void;
+};
+
+const TabButton: React.FC<TabButtonProps> = ({
+  children,
+  tabIndex,
+  selectedTabIndex,
+  setSelectedTabIndex,
+}) => (
+  <Button
+    variant="outlined"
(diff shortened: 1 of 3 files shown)


jadmnsaad:
Fixing inconsistent padding by changing colors? Is there another way?

djnunez-aot:
> Fixing inconsistent padding by changing colors? Is there another way?
@jadmsaadaot the problem is  we swap from an outlined button to a contained button on press originally and they have different styling. Maybe we could create a custom component and it does all that logic handling for us but i don't know if that's valuable

jadmnsaad:
> > Fixing inconsistent padding by changing colors? Is there another way?
> > @jadmsaadaot the problem is  we swap from an outlined button to a contained button on press originally and they have different styling. Maybe we could create a custom component and it does all that logic handling for us but i don't know if that's valuable

We should ideally stay away from spot changes to colors for atomic components like Buttons. We could perhaps update the styling in the theme. If we created a custom component for button we should probably use it everywhere. Or maybe just a spot change to the padding could be okay.

djnunez-aot:
> > > Fixing inconsistent padding by changing colors? Is there another way?
> > > @jadmsaadaot the problem is  we swap from an outlined button to a contained button on press originally and they have different styling. Maybe we could create a custom component and it does all that logic handling for us but i don't know if that's valuable
> 
> We should ideally stay away from spot changes to colors for atomic components like Buttons. We could perhaps update the styling in the theme. If we created a custom component for button we should probably use it everywhere. Or maybe just a spot change to the padding could be okay.

sounds good ill change the solution to do that instead then ill try to add a common component for this

jadmnsaad (approved):

Issue #1821, opened 2024-02-07
Insights - Dashboard


Connor1892:
**As an** Executive

**I want** a dashboard for my insights

**so that** I can navigate easily

### Scenario 1: 

GIVEN I am logged into EPIC.track
WHEN I select the Insights menu item
THEN I see a dashboard of all available Insights

### Validation Rules? 

(If yes, list here)

### Screenshot 

![MicrosoftTeams-image (3)](https://github.com/bcgov/EPIC.track/assets/126211473/3ef83552-53c2-4b7f-901c-184c9f068d3d)

(Link above will always be the most up-to-date version.)

maria-kuzmenko:
link to navigation - https://www.figma.com/file/dwlYoKngyXkuBoEGiEld1h/Design-System?type=design&node-id=975%3A1671&mode=design&t=kjWqhjVkm6E9fwj4-1

Connor1892:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/EPICtrack-63891ea941d309001fa292cf/issues/bcgov/EPIC.track/1821?planning-poker) with Zenhub @jadmsaadaot @dinesh-aot @djnunez-aot

PR #1976, opened 2024-03-11
Insights staff - Scrollable containers & Officer/analyst fetch

djnunez-aot:
#1821 

-Add scrollable containers

-Add fetch to get officer and analyst roles


epictrack-web/src/components/insights/Work/Tabs/Staff/Charts/workListing.tsx
```diff
@@ -29,22 +42,60 @@ const WorkList = () => {
...
+  }, []);
+
+  React.useEffect(() => {
+    getWorkStaffAllocation();
+  }, []);
+
+  React.useEffect(() => {
+    let columns: Array<MRT_ColumnDef<WorkOrWorkStaff>> = [];
+    if (wsData && wsData.length > 0) {
+      const rolename = "Officer / Analyst";
```
dinesh-aot: I think we have role id coming in the api response. It is always better to go with the ID cz the name might change in future.

djnunez-aot: good suggestion

Issue #1919, opened 2024-03-01
My Workplans - Title Truncation


Connor1892:
**Describe the bug**
when the screen size changes, everything on the work card gets truncated except for the title

**To Reproduce**
1. Go to my workplans
2. shrink the browser window horizontally
3. look at the title field

**Expected behavior**
the title should truncate nicely like the other fields

**Screenshots**

**Additional context**

Connor1892:
@maria-kuzmenko

maria-kuzmenko:
@djnunez-aot, the title should truncated but there should not be that much of white space. Can we set up fixed width between the title and the 'in progress' to 16px, so truncation will start when it reaches those margins 

<img width="1404" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNGZ0QXc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--4c77a0c35568fa1c73c9d8764f527e29ea51c2a3/Screen%20Shot%202024-03-18%20at%202.04.30%20PM.png" alt="Screen Shot 2024-03-18 at 2.04.30 PM.png" />

PR #1983, opened 2024-03-12
Add title truncation to work title

djnunez-aot:
#1919 

image

add title truncation to work title

### DIFF
--- a/epictrack-web/src/components/myWorkplans/Card/CardBody.tsx
+++ b/epictrack-web/src/components/myWorkplans/Card/CardBody.tsx
@@ -43,20 +43,21 @@ const CardBody = ({ workplan }: CardProps) => {
     >
       <Grid item container spacing={2}>
         <Grid item xs={9}>
-          <ETHeading4
-            bold
-            color={Palette.neutral.dark}
-            xs
-            sx={{
-              overflow: "hidden",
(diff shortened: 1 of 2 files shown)


epictrack-web/src/components/myWorkplans/Card/CardBody.tsx
```diff
@@ -43,20 +43,21 @@ const CardBody = ({ workplan }: CardProps) => {
...
-              overflow: "hidden",
-              textOverflow: "ellipsis",
-              whiteSpace: "nowrap",
-            }}
-          >
-            <Tooltip title={workTitle} placement="top-start">
-              <span>{workTitle}</span>
-            </Tooltip>
-          </ETHeading4>
+          <Tooltip title={workTitle} placement="top-start">
```
dinesh-aot: Did you check the tooltip. Is it working. As I understand the Tooltip component work only if it wraps a built in component not ETHeading4 which is custom component. Also, please check the way we have implemented enableTooltiop and enableEllipsis stuff in other components in shared folder

djnunez-aot: good idea i added the tooltop to enable the tool tip and ellipsis

Issue #1874, opened 2024-02-22
All Filters Expected Ordering


Connor1892:
---
**CARLY/CONNOR: please add more info to this card - like a table checklist of all pages with Filters/Tables/Filter Headers**

update TableFilter to have natural sort by default

the only exception should be Staff page in List management. (and only because the position filter has already been sorted hieratically)

---

**As a** user

**I want** Natural sorting to be the default on filters

**so that** I know where I can expect to find options in the dropdown

### Scenario 1: 

GIVEN I am interacting with any grid
WHEN I open a filter
THEN I see the options are sorted by natural sort

### Scenario 2:
 xxxxxx

### Validation Rules? 

(If yes, list here)

### Screenshot 

(Link above will always be the most up-to-date version.)

dinesh-aot:
@salabh-aot It would be good to have then naturally sorted from the api itself. The postgres doesn't allow that by default though we can try some custom functions to have this done. Lets discuss

dinesh-aot:
Here are the list of entities to be sorted in different places of epictrack
 - [x]  WorkForm - Responsible Ministry - Sort Order Column
 - [x]  WorkForm - Responsible EPD - Natural Sort
 - [x]  WorkForm - WorkLead - Natural Sort
 - [x]  WorkForm - Decision Maker - Natural Sort
 - [x]  ProponentForm - Relationship Holder - Natural Sort
 - [x]  ProjectForm - Proponents - Natural Sort

Connor1892:
we wanted the default for the filters to be natural sort with only a few exceptions. can it be build into the common component?

here are a few places we found wrongly ordered things:

insights:
- [x] lead - Natural
- [x] officer analyst - Natural
- [x] project - Natural
- [x] current phase - SortOrder of phase_codes table
- [x] other ministry (sort Order)
- [x] nations - Natural
- [x] proponents - Natural
- [x] subtype - Sort Order
- [ ] 

My Workplans
- [x] worktype (sort order)

maria-kuzmenko:
@djnunez-aot , @dinesh-aot 

sorting for Lead appears with duplication (too many Amy Avila). We can't do sorting by Analyst, cause of the non-existance of this column haha

[...]

PR #1990, opened 2024-03-14
All Filters Expected Ordering

salabh-aot:
#1874 - Added utility function to perform natural sort.


epictrack-api/src/api/resources/indigenous_nation.py
```diff
@@ -110,10 +111,10 @@ def get():
         indigenous_nations = IndigenousNationService.find_all_indigenous_nations(
             args.get("is_active")
         )
+        response = res.IndigenousResponseNationSchema(many=True).dump(indigenous_nations)
+        response = natural_sort(response, "name")
```
dinesh-aot, line 115: Shall we do this inside the service?

salabh-aot: The `natsort` library does not support sorting of objects. We either need to do `res.IndigenousResponseNationSchema(many=True).dump(indigenous_nations)` inside the service, or keep `natural_sort` call in the resource.

Issue #1975, opened 2024-03-11
Show Special history fields 30-60-90


dinesh-aot:
**Overview**
Look through the special history for the below fields for the given report date
- Project Name

Issue #1932, opened 2024-03-04
Referral Schedule - Staleness Chips Logic


Connor1892:
**Describe the bug**
any out of date (Red) chip says needs status instead of the last status date.

**To Reproduce**

**Expected behavior**
The Red Chip should still show the last status date (if there is one) on the work

**Screenshots**

**Additional context**

dinesh-aot:
@salabh-aot Even if the work has status updated in a very older date, it still shows "Needs Status" in the chip.

Connor1892:
@dinesh-aot 
do we need to change this logic to check against report date instead of today's date?
If we're looking at actually giving them the correct information for whatever date they choose, then we need to fix the referral schedule in other ways as well. (it shows a yellow status from the 29th when I call a report from the 30th while the real date is the 7th)

dinesh-aot:
**Help to test**
You can adjust the report_date to see the staleness changes. Previously it was based on today's date and now we have changed the staleness based on the report_date. So if you request RF based on an older date, we will check the last status updated date and older report_date to show the staleness.

Connor1892:
the colours are correct, and change correctly for each date chosen.

HOWEVER!!!
the colours are updating on date select instead of on submit.
this can lead to false view of a report on web view.

PR #2003, opened 2024-03-19
Show Special history fields 30-60-90

salabh-aot:
- Show Special history fields 30-60-90  
   - #1975 - Updated 30-60-90 report to show project name from special field history table, based on the date.
 - Updated project & proponent import APIs to save default special field history entries also.
 - Referral Schedule - Staleness Chips Logic
   - #1932 - Updated referal schedule report to use report date instead of current date for staleness logic.


epictrack-api/src/api/reports/thirty_sixty_ninety_report.py
```diff
@@ -159,16 +165,21 @@ def _format_data(self, data):
                 work["work_id"], work["anticipated_decision_date"]
             )
             event_decision_date = work["anticipated_decision_date"]
-            work[
-                "anticipated_decision_date"
-            ] = next_major_decision_event.anticipated_date
+            work["anticipated_decision_date"] = (
+                next_major_decision_event.anticipated_date
+            )
+            project_name = self._get_special_field_project_history(
```
dinesh-aot: This call would go in the for loop. Shall we make it a single call. If we have all the project ids collected prior and make a single call and do the replacement of project name in the works. I think we have the project ids set as index. I am not sure the IN clause would work effectively with indexes though.

epictrack-api/src/api/reports/thirty_sixty_ninety_report.py
```diff
@@ -404,3 +422,19 @@ def _get_table_data_and_styles(self, data, normal_style):
             )
             table_data.extend(period_data)
         return table_data, styles
+
+    def _get_special_field_project_history(self, project_id: int, date: datetime) -> str:
```
dinesh-aot: Also, can this kind of method be part of the special_field service.

epictrack-api/src/api/reports/thirty_sixty_ninety_report.py
salabh-aot: Updated to do batch queries for projects of 30, 60 & 90 days periods.

dinesh-aot (approved):

PR #2006, opened 2024-03-19
Make title truncation dependent on screen size

djnunez-aot:
#1919 

-Add media query to make the title truncation more consistent on different screens


epictrack-web/src/components/myWorkplans/Card/CardBody.tsx
```diff
@@ -51,7 +52,7 @@ const CardBody = ({ workplan }: CardProps) => {
             enableEllipsis
             tooltip={workTitle}
             sx={{
-              maxWidth: "70%",
+              maxWidth: isSmallScreen ? "75%" : "100%",
```
jadmnsaad: we can use sx={{xs:...

djnunez-aot: done

PR #2027, opened 2024-03-22
Restrict issue update post date past values

jadmnsaad:
https://app.zenhub.com/workspaces/epictrack-63891ea941d309001fa292cf/issues/gh/bcgov/epic.track/1729

Details:
- Add front-end restrictions on past values for issue update posted_date
- Add backend validation on issue update posted_date


davidjnunezf (approved):

epictrack-api/src/api/services/work_issues.py
```diff
@@ -169,3 +173,25 @@ def _check_edit_update_auth(cls, work_id, issue_update):
             authorisation.check_auth(one_of_roles=one_of_roles)
         else:
             cls._check_edit_auth(work_id)
+
+    @classmethod
+    def _check_update_date_validity(cls, work_issue, update_data, issue_update_id=None):
+        """Check if edited date is valid"""
+        if update_data.get('posted_date').timestamp() < work_issue.start_date.timestamp():
```
dinesh-aot, line 180: It would be good to compare the date only. Here it will consider upto the millisecond. Please check.

jadmnsaad: It was like that initially. But the tests started failing with an error that said TypeError: can't compare offset-naive and offset-aware datetimes. One of the suggested solutions I found is to convert it to timestamp. Any other suggested solutions?

PR #2032, opened 2024-03-25
Fetch only active team lead staff and prevent lead duplication

jadmnsaad:
https://app.zenhub.com/workspaces/epictrack-63891ea941d309001fa292cf/issues/gh/bcgov/epic.track/2004

Details:
- Add a backend check to prevent team lead duplication
- Only fetch active staff for work staff info in Work Cards

### DIFF
--- a/epictrack-api/src/api/models/staff_work_role.py
+++ b/epictrack-api/src/api/models/staff_work_role.py
@@ -51,7 +51,7 @@ class StaffWorkRole(BaseModelVersioned):
     @classmethod
     def find_by_role_and_work(cls, work_id: int, role_id):
         """Return by work and role ids."""
-        return cls.query(StaffWorkRole).filter(
+        return cls.query.filter(
             StaffWorkRole.work_id == work_id,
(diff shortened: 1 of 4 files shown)


epictrack-api/src/api/models/staff_work_role.py
```diff
@@ -47,3 +47,28 @@ def as_dict(self):  # pylint:disable=arguments-differ
     def find_by_work_id(cls, work_id: int):
         """Return by work id."""
         return cls.query.filter_by(work_id=work_id)
+
+    @classmethod
+    def find_by_role_and_work(cls, work_id: int, role_id):
```
dinesh-aot, line 52: would it be good to take the first matched result where this function is being used. This can act as a function which returns the StaffWorkRole for a given work and role id.

jadmnsaad: Okay, I will remove first() and do all() instead

dinesh-aot (approved):

PR #2067, opened 2024-04-02
Connect Floating help button to help pages

jadmnsaad:
https://app.zenhub.com/workspaces/epictrack-63891ea941d309001fa292cf/issues/gh/bcgov/epic.track/2035

Details:
- Add SupportCenterMenuItem
- Add help pages map  json
- Add algorithm to choose best match
- Disabled scroll lock to prevent it from stopping page scroll and subsequently moving when opened in chrome where the scroll bar is always visible


epictrack-web/src/components/AppHelpButton/SupportCenterMenuItem.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+type HelpPageLink = {
+  epicTrackPath: string;
+  helpPage: string;
+};
+
+const SupportCenterMenuItem = () => {
+  const handleClick = () => {
+    const currentPath = window.location.pathname.split("?")[0]; // Get pathname without query params
+    const currentPathSegments = currentPath.split("/"); // Split current path
```
dinesh-aot, line 20: we might not get this path parameter for selecting the help page every time. For instance, if we are on work plan for Project Notification, we should be pointing to the help page which talks about Project Notification where the URL would be https://epictrack-dev.apps.gold.devops.gov.bc.ca/work-plan?work_id=58
This way would fit for regular pages.

epictrack-web/src/components/AppHelpButton/HelpMenuItem.tsx
```diff
@@ -0,0 +1,6 @@
+import { MenuItem, MenuItemProps } from "@mui/material";
+import { Palette } from "styles/theme";
```
dinesh-aot: The help button moves when you click. Please check if that is already handled in this PR

epictrack-web/src/components/AppHelpButton/SupportCenterMenuItem.tsx
jadmnsaad: how would you figure from this https://epictrack-dev.apps.gold.devops.gov.bc.ca/work-plan?work_id=58 that it should be directed to Project Notification help?

epictrack-web/src/components/AppHelpButton/HelpMenuItem.tsx
jadmnsaad: hmm interesting, It's happening in chrome. I use firefox and it's not happening there, I will look into it

djnunez-aot (approved):

Issue #1964, opened 2024-03-08
Project change in work doesn't reflect in the existing work titles


dinesh-aot:
**Overview**
If you change the name of the project associated with a work and then go to the work plan. The changed name won't be visible in the title of the work.

**Expectation**
The project change should be reflected in the work title.

**Technical Findings**
The title column in work is currently a persisted field which can be a derived one since all the part of the field is derived from existing data.

PR #2075, opened 2024-04-04
removed title column and added a generated property

saravanpa-aot:
https://github.com/bcgov/EPIC.track/issues/1964

1. removed title and replaced it with a property
2. Addressed code to check the duplicate.
3. test cases modified


epictrack-api/src/api/models/work.py
```diff
@@ -108,14 +115,27 @@ def as_dict(self, recursive=True):
     @classmethod
     def check_existence(cls, title, work_id=None):
         """Checks if a work exists for a given title"""
-        query = Work.query.filter(
-            func.lower(Work.title) == func.lower(title), Work.is_deleted.is_(False)
-        )
+        title_parts = title.split()
+        project_type_name, work_type_name, simple_title = (title_parts[i].strip() if i < len(title_parts)
+                                                           else None for i in range(3))
```
jadmnsaad, line 120: could the code styling be improved here?

epictrack-api/src/api/models/work.py
```diff
@@ -108,14 +115,27 @@ def as_dict(self, recursive=True):
...
+        # pylint: disable=import-outside-toplevel
+        from api.models.work_type import WorkType
```
jadmnsaad, line 122: can we not keep imports at the top?

epictrack-api/src/api/utils/util.py
```diff
@@ -45,3 +45,11 @@ def find_index_in_array(json_array, target_object):
...
             return index
     return -1  # Return -1 if not found
+
+
+def generate_title(project_name, work_type_name, simple_title=''):
+    """Generate the title string for a work."""
+    parts = [project_name, work_type_name]
+    if simple_title:
+        parts.append(simple_title)
+    return ' - '.join(parts)
```
jadmnsaad, line 55: this will create coupling with the front-end, there it is separated by - and now here it is as well. Are we okay with this?

epictrack-api/src/api/services/work.py
```diff
@@ -280,6 +280,15 @@ def create_work(cls, payload, commit: bool = True):
             db.session.commit()
         return work
 
+    @classmethod
+    def _check_duplicate_title(cls, payload, work_id=None):
+        """Check if the title exists."""
+        project = Project.find_by_id(payload.get('project_id'))
+        work_type: WorkType = WorkType.find_by_id(payload.get('work_type_id'))
+        title = util.generate_title(project.name, work_type.name, payload.get('simple_title'))
```
jadmnsaad: if title is a property now, why are we generating it again manually?

jadmnsaad (changes requested):
Added few comments, also please description

epictrack-api/src/api/models/work.py
saravanpa-aot: any suggestions ? you meant to make it more readable ?

epictrack-api/src/api/models/work.py
saravanpa-aot: nope..circular dependency

epictrack-api/src/api/utils/util.py
saravanpa-aot: yeah.. this is by design..

epictrack-api/src/api/services/work.py
saravanpa-aot: i just improved variable name and added a comment as well..plz see if that improves

epictrack-api/src/api/models/work.py
saravanpa-aot: first version was individual variables , but felt too boring ..so went for this..but definitly welcoming all suggestions here..

jadmnsaad: Oh no sorry, it's just the styling I meant, the indentation felt a bit weird for me but I guess it just fits Python styling conventions. No need to change, unless you feel it's easy or better

jadmnsaad (approved):

Issue #2059, opened 2024-03-28
Special History - Size problems


Connor1892:
**Describe the bug**
the Special history dropdown in the project modal is cut to half the modal size. this makes it very hard to use

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

Connor1892:
on Edit Project Modal, if I open both both Name and Proponent, I can type and change the name and lock it again when locking the proponent again.

opening the Name pushes the proponent down to the next line.

we should not be able to have both special histories at the same time.

djnunez-aot:
@Connor1892  the name pushing down the proponent is something i discussed with Jad, it would take a very large redesign of the component might not be worth our efforts. Working on the rest of the changes.

Connor1892:
all fields should be disabled when special history is open.
see Edit Work modal as example.

PR #2078, opened 2024-04-08
Fix Special Field size on pop up

djnunez-aot:
#2059 

-Fix special field dropdown size

### DIFF
--- a/epictrack-web/src/components/shared/specialField/index.tsx
+++ b/epictrack-web/src/components/shared/specialField/index.tsx
@@ -312,7 +312,6 @@ export const SpecialFieldGrid = ({
         background: Palette.white,
         overflowX: "auto",
         position: "absolute",
-        zIndex: 2,
       }}
     >
       <Box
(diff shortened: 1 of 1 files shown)


epictrack-web/src/components/shared/specialField/index.tsx
```diff
@@ -311,6 +311,8 @@ export const SpecialFieldGrid = ({
         border: `2px solid ${Palette.neutral.bg.dark}`,
         background: Palette.white,
         overflowX: "auto",
+        position: "absolute",
+        zIndex: 2,
```
jadmnsaad: why do we need zIndex? can we remove it?

Also zIndex is better obtained from the mui theme

const theme = useTheme()

zIndex: theme.zIndex...

jadmnsaad (changes requested):
Just a concern regarding zIndex

epictrack-web/src/components/shared/specialField/index.tsx
djnunez-aot: not needed will remove

PR #2093, opened 2024-04-09
Fix special field grid pop up

djnunez-aot:
#2078 
-for specialfield: some textfields cover other fields adding z index so it can pop up at the highest layer


epictrack-web/src/components/shared/specialField/index.tsx
```diff
@@ -312,6 +313,7 @@ export const SpecialFieldGrid = ({
         background: Palette.white,
         overflowX: "auto",
         position: "absolute",
+        zIndex: theme.zIndex.modal + 1,
```
jadmnsaad: why are adding this?

djnunez-aot: its being blocked by other components on test env

dinesh-aot (approved):

Issue #2071, opened 2024-04-03
Confetty shows false completion celebrations


dinesh-aot:
**Describe the bug**
Confetty shows false completion celebrations when you choose a decision to create more events in the same phase and make on of them the new END event.

**To Reproduce**
1. Go to All Works
2. Create a new assessment work
3. Go to workplan of the created work
4. fill the actual dates in the first phase until the end event(the event in the last position in bold)
5. When you put actual date in the last event, choose "CEAO requires the proponent to submit revised IPD/EP"
6. This decision will cause the back end to change the current end event to an INTERMEDIATE event and add two more events to the phase and make one of them the new end event

**Expected behavior**
App should wait to show confetty for the phase to be actually completed by putting the actual event on the newly created end event

**Screenshots**

**Additional context**

PR #2101, opened 2024-04-10
Confetti fix - add get by work phase Id route

djnunez-aot:
#2071 

-Add getByWorkPhaseId endpoint 

-Refactor getTemplate endpoint


epictrack-web/src/components/workPlan/event/EventList.tsx
```diff
@@ -495,6 +484,17 @@ const EventList = () => {
     }
   }, [ctx.selectedWorkPhase?.work_phase.phase.id]);
 
+  const getWorkPhaseById = React.useCallback(async () => {
```
dinesh-aot, line 505: I hope this should be invoked only if the event.event_position == END

djnunez-aot: will do that

PR #2132, opened 2024-04-17
Refactor ETChip - add booleans to control styling

djnunez-aot:
-Change et chips to have booleans passed down to control styling

### DIFF
--- a/epictrack-web/src/components/indigenousNation/IndigenousNationList.tsx
+++ b/epictrack-web/src/components/indigenousNation/IndigenousNationList.tsx
@@ -9,7 +9,7 @@ import IndigenousNationForm from "./IndigenousNationForm";
 import { Staff } from "../../models/staff";
 import staffService from "../../services/staffService/staffService";
 import { MasterContext } from "../shared/MasterContext";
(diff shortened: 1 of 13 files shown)


epictrack-web/src/components/indigenousNation/IndigenousNationList.tsx
```diff
@@ -216,12 +216,8 @@ export default function IndigenousNationList() {
         },
         Cell: ({ cell }) => (
           <span>
-            {cell.getValue<boolean>() && (
-              <ActiveChip label="Active" color="primary" />
-            )}
-            {!cell.getValue<boolean>() && (
-              <InactiveChip label="Inactive" color="error" />
-            )}
+            {cell.getValue<boolean>() && <ETChip active label="Active" />}
```
jadmsaadaot, line 219: What do you think of using a variant prop like in MUI. The boolean will only work for two cases active or inactive. A variant that takes a string "active", "inactive", "somethingelse" would allow us to add more chip variants

jadmnsaad (approved):
Consider using variant instead of a

Issue #2104, opened 2024-04-10
Insights Page - Hoverstates


Connor1892:
**Describe the bug**
we want the hover to be the same:
- [x] staff: work name should be clickable
- [x] staff: hover needed on truncated text
- [x] staff: team & lead should have fixed width columns (same as on all works)
- [x] staff: officer Analyst should not be blue
- [x] general: work name should be clickable
- [x] general: hover needed on truncated text
- [x] partners: work name should be clickable
- [x] partners: hover needed on truncated text
- [x] partners: hoverstate for nations column needs to 'return' seperated (like a bullet list) instead of comma seperated
- [x] partners: federal involvement should have fixed width columns (same as on all works)

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

Connor1892:
@djnunez-aot 
None of the fixed width columns are fixed width.
the nations column needs to be in-line, only the hoverstate should be bulleted.

PR #2154, opened 2024-04-23
Insights Page, Hover states and onclicks for work names

djnunez-aot:
#2104 

Made the following changes:
    staff: work name should be clickable
    staff: hover needed on truncated text
    staff: team & lead should have fixed width columns (same as on all works)
    staff: officer Analyst should not be blue
    general: work name should be clickable
    general: hover needed on truncated text
    partners: work name should be clickable
    partners: hover needed on truncated text
    partners: hoverstate for nations column needs to 'return' separated (like a bullet list) instead of comma separated
    partners: federal involvement should have fixed width columns (same as on all works)

### DIFF
--- a/epictrack-web/src/components/insights/Work/Tabs/Staff/Charts/workListing.tsx
+++ b/epictrack-web/src/components/insights/Work/Tabs/Staff/Charts/workListing.tsx
@@ -12,21 +12,11 @@ import MasterTrackTable from "components/shared/MasterTrackTable";
 import { WorkStaff } from "models/workStaff";
 import { Role, WorkStaffRole, WorkStaffRoleNames } from "models/role";
 import { useGetWorkStaffsQuery } from "services/rtkQuery/workStaffInsights";
-import { All_WORKS_FILTERS_CACHE_KEY } from "components/work/constants";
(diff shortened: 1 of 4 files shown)


epictrack-web/src/components/insights/Work/Tabs/Staff/Charts/workListing.tsx
```diff
@@ -169,33 +163,31 @@ const WorkList = () => {
...
+        size: 200,
+        Cell: ({ row, renderedCellValue }) => {
+          return (
+            <ETGridTitle
+              to={`/work-plan?work_id=${row.original.work.id}`}
+              titleText={row.original.work.title}
+              enableTooltip
+              tooltip={row.original.work.title}
+            >
+              {row.original.work.title}
```
dinesh-aot: Why did we remove the {renderedCellValue}. This is used to show the search highlight on the title in the grid. It was done per request before. Could you please check it with Maria

djnunez-aot: reverted row.original.work.title i did this originally because i was debugging why the name was not displaying on the chart

Issue #1802, opened 2024-02-02
Grid Download buttons


Connor1892:
**As a**  user

**I want** to be able to download any grid in track

**so that** I can easily export data

### Scenario 1: 

GIVEN I am on a page that has a grid
WHEN I select the download button
THEN I receive an excel document **with only the data I've filtered down to**

### Scenario 2:
 xxxxxx

### Validation Rules? 

(If yes, list here)

### Screenshot 

(Link above will always be the most up-to-date version.)

Connor1892:
TODO: find all pages with grids

Connor1892:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/EPICtrack-63891ea941d309001fa292cf/issues/bcgov/EPIC.track/1802?planning-poker) with Zenhub @dinesh-aot @djnunez-aot @jadmsaadaot

Connor1892:
@djnunez-aot 
missing on:
- project dashboard
- admin users
- team tab (should match nation tab)

broken on:
- resource forecast (no name field in csv)

Connor1892:
can't test the resource forecast CSV download because resource forecast is not showing anything

maria-kuzmenko:
@djnunez-aot  can we make the consistency with download buttons and reuse the one we have in the workplan grid? Thanks!
<img width="1679" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOTNHQkE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--2f811cfafeaa331954e8f8cdd27972f22ba7f33e/Screen%20Shot%202024-05-08%20at%204.09.07%20PM.png" alt="Screen Shot 2024-05-08 at 4.09.07 PM.png" />

PR #2184, opened 2024-04-30
Setup Exporting to csv for all grids

djnunez-aot:
#1802 

-Add download csv to all grids

-creating a reusable exporting function for all master track tables to use


epictrack-web/src/utils/exportUtils.ts
```diff
@@ -0,0 +1,37 @@
...
+import dateUtils from "./dateUtils";
+import { MRT_TableInstance } from "material-react-table";
+import { MRT_RowData } from "material-react-table";
+interface ExportToCsvOptions<T extends MRT_RowData> {
+  table: MRT_TableInstance<T>;
+  downloadDate: string | null;
+  filenamePrefix: string;
+}
+
+export async function exportToCsv<T extends MRT_RowData>({
```
jadmnsaad: This seems to be specific to the material react table.

let's either:
- Make it more general so that any component can use it
- Place it under the same folder where the track table is

epictrack-web/src/components/proponent/ProponentList.tsx
```diff
@@ -255,6 +266,19 @@ export default function ProponentList() {
                     Create Proponent
                   </Button>
                 </Restricted>
+                <Tooltip title="Export to csv">
+                  <IconButton
+                    onClick={() =>
+                      exportToCsv({
+                        table,
+                        downloadDate: new Date().toISOString(),
+                        filenamePrefix: "proponent-listing",
```
dinesh-aot, line 275: Can you give me a screen shot of how this would look in the table

djnunez-aot: this is using a csv visualizer 
image

epictrack-web/src/utils/exportUtils.ts
djnunez-aot: moved the export function under the table utils

Issue #2162, opened 2024-04-25
Admin Page - User Roles should be in order of access


Connor1892:
---

reorder the edit roles modal dropdown to have Dev at the top, and viewer at the bottom.
- Dev
- System Admin
- Instance Admin
- Superuser
- Viewer
- Blank

---

**As a** (type of user)

**I want** (goal)

**so that** (reason/ why/ value)

### Scenario 1: 

GIVEN ... (you or your condition)
WHEN ... (what you do)
THEN ... (what you see)

### Scenario 2:
 xxxxxx

### Validation Rules? 

(If yes, list here)

### Screenshot 

(Link above will always be the most up-to-date version.)

Connor1892:
@djnunez-aot 
my bad.
the users grid should be sorted alphabetically, this card is supposed to be about the dropdown in the edit modal.

PR #2186, opened 2024-05-01
On user table load sort by group level initially

djnunez-aot:
#2162 

-Reorder in terms of group level


jadmnsaad (approved):

epictrack-web/src/components/user/UserList.tsx
```diff
@@ -36,7 +36,12 @@ const UserList = () => {
     try {
       const userResult = await UserService.getUsers();
       if (userResult.status === 200) {
-        setUsers(userResult.data as never);
+        const sortedResults = (userResult.data as User[]).sort((a, b) => {
```
dinesh-aot, line 40: Do you really need Infinity here. Can't it just work with zero

djnunez-aot: we can do 0

Issue #2176, opened 2024-04-29
Work Staff - fix grid


Connor1892:
---
remove second EPD column
fix work name column

---

**As a** (type of user)

**I want** (goal)

**so that** (reason/ why/ value)

### Scenario 1: 

GIVEN ... (you or your condition)
WHEN ... (what you do)
THEN ... (what you see)

### Scenario 2:
 xxxxxx

### Validation Rules? 

(If yes, list here)

### Screenshot 

(Link above will always be the most up-to-date version.)

Connor1892:
@djnunez-aot 
we need hover states for truncated text

dinesh-aot:
@djnunez-aot  Can you move the work_title column to the first index in the grid

Connor1892:
@djnunez-aot 
still looking for hoverstates on truncated test. (ie: officer/analyst, project, etc.)

Connor1892:
@djnunez-aot 
can't test.
Work-staff page is not showing any results.

PR #2187, opened 2024-05-02
Remove responsible epd and add work title to work staff table

djnunez-aot:
#2176 

-remove responsible epd column

-Add work title to the work staff result


jadmnsaad (changes requested):
Can we not add the necessary column in backend?

Issue #2183, opened 2024-04-30
Workplan Page - Current Phase not Registering


Connor1892:
**Describe the bug**
when using the milestone to delete intake of Minister's designation, the new "first phase" does not register as the current phase.

this may have other implications behind the scenes.

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

Connor1892:
needs to have a page refresh before it registers. is this part of this card?

Connor1892:
also does this need an estimate or is it the same as #2071

Connor1892:
> needs to have a page refresh before it registers. is this part of this card?

@dinesh-aot

Issue #2016, opened 2024-03-20
Insights - Work Dashboard - Staff - Count Bug


carlygreenehill:
**Describe the bug**
From the Staff Insights page, I can see counts for Work by Lead and Work by Staff. However, when I go to filter by Lead or Officer/Analyst in the grid below, the # of results come back differently. 

Related to #1852

This is likely related to Scenario 3 & 4 - let's have a chat at DAILY SCRUM on Thurs March 21 

**To Reproduce**

1. Go to Insights Tab 
2. Click on the Staff page
3. Scroll down to the Grid to filter by Officer/Analyst or Team Lead
4. See counts differ between Grid and Graphs

**Expected behavior**
Same counts and no Team Leads counted toward Officer/Analyst and vice versa

**Screenshots**

**Additional context**

---

SCENARIO 2 - Grid is pulling inactive works in GENERAL

Connor1892:
when searching for jad in the officer/ analyst column, it shows 0 results. but the graphic claims he's on 3 works. when you check the table he's in there with other staff.
the filter should pick up when he's in the cell, not just when it's ONLY him in the cell.

Connor1892:
also we cannot confirm the staff showing are correct, because the names of the works are all duplicates. not sure which card introduced this.

PR #2195, opened 2024-05-04
Insight result inconsistencies

dinesh-aot:
#2016 
#2183


epictrack-api/src/api/insights/work_lead_insight.py
```diff
@@ -18,17 +20,23 @@ def generate_partition_query(self):
         """Generates the group by subquery."""
         partition_query = (
             db.session.query(
-                Work.work_lead_id,
+                StaffWorkRole.staff_id,
                 func.count()
-                .over(order_by=Work.work_lead_id, partition_by=Work.work_lead_id)
+                # .over(order_by=Work.work_lead_id, partition_by=Work.work_lead_id)
```
jadmnsaad: if we are not using this should we remove it? YAGNI

epictrack-api/src/api/insights/work_staff_insight.py
```diff
@@ -20,17 +21,18 @@ def generate_partition_query(self):
             db.session.query(
                 StaffWorkRole.staff_id,
                 func.count()
-                .over(
-                    order_by=StaffWorkRole.staff_id, partition_by=StaffWorkRole.staff_id
-                )
+                # .over(
+                #     order_by=StaffWorkRole.staff_id, partition_by=StaffWorkRole.staff_id
+                # )
```
jadmnsaad: should remove this since we commented it out since YAGNI

epictrack-api/src/api/resources/work.py
```diff
@@ -128,7 +128,10 @@ class WorkResources(Resource):
     @profiletime
     def get():
         """Return all resource and work details"""
-        works = WorkService.find_allocated_resources()
+        is_active = request.args.get('is_active')
+        if is_active is not None:
+            is_active = is_active.lower() == 'true'
```
jadmnsaad: do we need this if statement.

can we use this:
image

or just args.get(is_active, False, bool)?

epictrack-web/src/components/shared/MasterTrackTable/index.tsx
```diff
@@ -225,7 +225,7 @@ const MasterTrackTable = <TData extends MRT_RowData>({
         return filterValue.includes(row.getValue(id));
       },
     },
-    ...otherPropsData,
+    ...otherProps,
```
jadmnsaad: unstable reference, will cause infinite rerender

jadmnsaad: https://www.material-react-table.com/docs/guides/memoization#what-to-memoize

epictrack-web/src/utils/index.ts
```diff
@@ -6,20 +6,16 @@ const sort = (collection: any[], sortField: string) => {
...
+  if (keys && keys.length === 0) {
+    return collection.sort(function (a, b) {
+      return collator.compare(a[sortField], b[sortField]);
     });
-
-    return collator.compare(aValue, bValue);
-  });
+  } else {
+    return collection.sort(function (a, b) {
+      return collator.compare(a[keys[0]][keys[1]], b[keys[0]][keys[1]]);
```
jadmnsaad, line 16: nice, this works for our current needs. But I feel something like "key1.key2.key3" might break this, so maybe add a safety if statement if key.length > 2 for now and return something default

jadmnsaad: this would be a technical debt to be solved later maybe, by recursion probably

epictrack-api/src/api/resources/work.py
dinesh-aot: I used to do like this
image

epictrack-web/src/utils/index.ts
dinesh-aot: Lets assume we would go only two nested levels in the json for now

Issue #2180, opened 2024-04-30
Milestones - Phase End Milestone Anticipated


Connor1892:
---

because the anticipated date for phase end milestones cannot be moved, we should disable it.

---

**As a** (type of user)

**I want** (goal)

**so that** (reason/ why/ value)

### Scenario 1: 

GIVEN ... (you or your condition)
WHEN ... (what you do)
THEN ... (what you see)

### Scenario 2:
 xxxxxx

### Validation Rules? 

(If yes, list here)

### Screenshot 

(Link above will always be the most up-to-date version.)

Connor1892:
readiness decision is not greyed out, and lets us extend that timeline with out an extension indefinitely

PR #2199, opened 2024-05-06
Milestones - disabled anticipated date picker on phase end milestone

djnunez-aot:
#2180 

-Disable anticipated date on phase end


epictrack-web/src/components/workPlan/event/EventForm.tsx
```diff
@@ -679,6 +679,10 @@ const EventForm = ({
               <ETFormLabel required>{anticipatedLabel}</ETFormLabel>
               <ControlledDatePicker
                 name="anticipated_date"
+                disabled={Boolean(
+                  selectedConfiguration?.id &&
```
dinesh-aot: One more condition is required if you want to disable the datepicker. The workphase should be legislated

dinesh-aot (approved):

PR #2243, opened 2024-05-15
insights->filtering

dinesh-aot:
#1874

### DIFF
--- a/epictrack-web/src/components/insights/Work/Tabs/Partners/charts/WorkList.tsx
+++ b/epictrack-web/src/components/insights/Work/Tabs/Partners/charts/WorkList.tsx
@@ -38,9 +38,19 @@ const WorkList = () => {
       showNotification("Error fetching works", { type: "error" });
     }
   }, [error]);
-
   const federalInvolvements = useMemo(() => {
-    return Array.from(new Set(works.map((w) => w?.federal_involvement?.name)));
+    return Array.from(
+      new Set(
+        [...works]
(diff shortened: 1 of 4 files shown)


epictrack-web/src/components/insights/Work/Tabs/Partners/charts/WorkList.tsx
```diff
@@ -38,9 +38,19 @@ const WorkList = () => {
       showNotification("Error fetching works", { type: "error" });
     }
   }, [error]);
-
   const federalInvolvements = useMemo(() => {
-    return Array.from(new Set(works.map((w) => w?.federal_involvement?.name)));
+    return Array.from(
+      new Set(
+        [...works]
```
jadmnsaad, line 44: why the spread operator? why not just works instead of [...works].

dinesh-aot: The sort directly tries to edit the variable works and causes error "Cannot assign to read only property '0' of object '[object Array]'". Spread is required to create a copy and perform sort

jadmsaadaot: oh cool, fair enough

jadmnsaad: Ok I see, fair

PR #2244, opened 2024-05-15
Added last updated time field and excluded it from history updates

saravanpa-aot:
1) Added new column last_active_at to capture when was the last time a staff logged in
                - Added the new API
                - Added the rest call when keycloak init is done (basically login or hard refreshes)
                - Added alembic

2) Added a exclude_from_tracking 
                - Since this field doesnt need to be tracked , added the changes for not perisisting this fields value and also if this is the only field getting updated , skip the whole update
                - the only thing pending is we still need a dummy column in the history table though it wont have any values .Couldnt make it working.

### DIFF
--- a/epictrack-api/src/api/resources/staff.py
+++ b/epictrack-api/src/api/resources/staff.py
@@ -24,7 +24,6 @@ from api.services import StaffService
 from api.utils import auth, profiletime
 from api.utils.util import cors_preflight
 
-
 API = Namespace("staffs", description="Staffs")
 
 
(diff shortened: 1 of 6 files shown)


epictrack-api/src/api/resources/staff.py
```diff
@@ -100,6 +99,22 @@ def delete(staff_id):
...
 
 
+@cors_preflight('PUT')
+@API.route('/<int:staff_id>/last_active_at', methods=['PUT', 'OPTIONS'])
+class StaffLastActiveAt(Resource):
+    """Endpoint resource to return staff last active time."""
+
+    @staticmethod
+    @cors.crossdomain(origin='*')
+    @auth.require
```
dinesh-aot, line 109: usual question. This end point will ever update the whole object. Should this be PATCH instead of PUT

epictrack-api/src/api/services/staff.py
```diff
@@ -79,6 +79,14 @@ def update_staff(cls, staff_id: int, payload: dict):
         staff = staff.update(payload)
         return staff
 
+    @classmethod
+    def update_last_active(cls, staff_id: int):
+        """Update staff's last active time."""
+        staff = Staff.find_by_id(staff_id)
+        staff.last_active_at = datetime.now()
```
dinesh-aot, line 86: Can you check the base_mode.py. we are updateing the utc in both created_at and updated_at

saravanpa-aot: done

epictrack-api/src/api/resources/staff.py
saravanpa-aot: done

Issue #2201, opened 2024-05-06
My Tasks - Task Modals


Connor1892:
**As a** user

**I want** a place where all my tasks from all works are in one place

**so that** I can stay on top of my work

### Scenario 1: 

GIVEN I am on the My Tasks page
WHEN I click on the active task title
THEN I see the Task modal for that task

### Screenshot 1
<img width="1297" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNGU1QkE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--e418cd4357f4cdade89d3d9693b2c1825190975e/Screen%20Shot%202024-05-06%20at%2010.46.49%20AM.png" alt="Screen Shot 2024-05-06 at 10.46.49 AM.png" />

### Scenario 2:

GIVEN I have a Task Modal open
WHEN I make edits to the fields
THEN the only options are the ones that would show if I had opened the modal from that work's workplan

### Scenario 3:

GIVEN I have a Task Modal open
WHEN I make edits to the fields
THEN the edits I save will alter the task in the workplan

maria-kuzmenko:
link to prototype - https://www.figma.com/file/XcYynwH7xrO2P30uqewaLS/For-Development?type=design&node-id=128-1031&mode=design

maria-kuzmenko:
@djnunez-aot 
can you, please, fix the size of the modal - same as we use when you open task modal in the workplan grid?
See compared screenshots for the reference. 

Task Modal in Workplan grid
<img width="498" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOElPQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--2a1d467eedb218faf2552cb94e42c8417fe1d2db/Screen%20Shot%202024-05-24%20at%2010.16.10%20AM.png" alt="Screen Shot 2024-05-24 at 10.16.10 AM.png" />

Task Modal in Tasks
[...]

PR #2290, opened 2024-05-27
My Tasks Modal - remove form padding

djnunez-aot:
#2201 

-Remove padding on my task modal

### DIFF
--- a/epictrack-web/src/components/myTasks/MyTaskDialog.tsx
+++ b/epictrack-web/src/components/myTasks/MyTaskDialog.tsx
@@ -42,7 +42,7 @@ export const MyTaskDialog = ({ data }: MyTaskData) => {
       }}
       disableEscapeKeyDown
       fullWidth
-      maxWidth="lg"
+      maxWidth="md"
       okButtonText="Save"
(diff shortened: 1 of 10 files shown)


epictrack-web/src/components/myTasks/MyTaskDialog.tsx
```diff
@@ -54,6 +54,11 @@ export const MyTaskDialog = ({ data }: MyTaskData) => {
       saveButtonProps={{
         disabled: disableSave,
       }}
+      sx={{
+        "& .MuiDialogContent-root": {
+          padding: 0,
```
jadmnsaad: not great, we shouldn't overwrite mui classes like this. Find another way

jadmnsaad (changes requested):
We shouldn't overwrite mui classes like this

epictrack-web/src/components/myTasks/MyTaskDialog.tsx
djnunez-aot: i added a prop since this happens everywhere

jadmnsaad (approved):

Issue #2236, opened 2024-05-13
Work-staff Page - Download broken


Connor1892:
**Describe the bug**
when downloading the grid from the work-staff page, it doesn't have any data in the co-lead, officer/analyst, or work lead columns.

same problem found in the Staff Insights csv.
and Partners Insights csv missing nations.
and Workplan Teams Tab csv is missing Role.

we think it's just that the query isn't looking for these as they're newest columns

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

PR #2305, opened 2024-06-03
Download CSV bug - handle fields with no accessor keys

djnunez-aot:
#2236 

-Allow columns without accesor keys to access their data

### DIFF
--- a/epictrack-web/src/components/shared/MasterTrackTable/utils.ts
+++ b/epictrack-web/src/components/shared/MasterTrackTable/utils.ts
@@ -72,31 +72,6 @@ interface ExportToCsvOptions<T extends MRT_RowData> {
   filenamePrefix: string;
 }
 
-function getStaffNamesByRole(row: any, roleName: string): string {
-  return row.staff
-    .filter(
-      (staffMember: { role: { name: string } }) =>
-        staffMember.role.name === roleName
-    )
-    .map(
-      (staffMember: { first_name: string; last_name: string }) =>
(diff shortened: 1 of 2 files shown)


epictrack-web/src/components/shared/MasterTrackTable/utils.ts
```diff
@@ -72,17 +72,51 @@ interface ExportToCsvOptions<T extends MRT_RowData> {
...
+    .map(
+      (staffMember: { first_name: string; last_name: string }) =>
+        `${staffMember.last_name} ${staffMember.first_name}`
+    )
+    .join("; ");
+}
+
+const customAccessors: { [key: string]: (row: any) => any } = {
+  "Responsible EPD": (row) =>
+    `${row.responsible_epd?.first_name} ${row.responsible_epd?.last_name}`,
```
jadmnsaad: This is a utils file for the MasterTrackTable. Why do we have mention of Responsible EPD here and Work Lead here?

djnunez-aot: lots of tables dont have accessor keys that are common and it causes issues with exporting

jadmnsaad: but this component is general, it doesn't know or care about business logic like Responsible EPD or Work Lead. If this is the solution it shouldn't be here

djnunez-aot: should i just allow passing custom accessors then into the export function

PR #2332, opened 2024-06-14
delete function change pg

dinesh-aot:


epictrack-api/migrations/versions/f554f142190f_delete_work_change.py
```diff
@@ -0,0 +1,87 @@
...
+    $function$
+    ;
+    """)
+
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    with op.batch_alter_table('works_history', schema=None) as batch_op:
```
jadmnsaad: Are these downgrades related?

Issue #2255, opened 2024-05-17
Link between staff and idir user


dinesh-aot:
**Overview**
Staff are the primary users of Track application. All the staff users are supposed to have an IDIR prior to sign in the Track.
**Things to do**
1. Create a column idir_user_id in staff table
2. Copy the provider_id from keycloak based on the staff email address
3. Need a migration to fix the idir_user_id for the existing staff members

PR #2361, opened 2024-07-17
2255 - Add idir to staff model

djnunez-aot:
#2255


epictrack-api/migrations/versions/19227722dffc_set_up_idir.py
```diff
@@ -0,0 +1,39 @@
...
+
+    with op.batch_alter_table('staffs_history', schema=None) as batch_op:
+        batch_op.add_column(sa.Column('idir_user_id', sa.String(length=100), autoincrement=False))
+
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    with op.batch_alter_table('staffs_history', schema=None) as batch_op:
```
dinesh-aot: You can basically run the script along with this migration.

epictrack-api/src/api/services/staff.py
```diff
@@ -59,9 +59,19 @@ def find_all_non_deleted_staff(cls, is_active=False):
...
+        # Normalize the email and check for existence in the local database
+        email = payload["email"].lower()
+        exists = cls.check_existence(email)
+
         if exists:
             raise ResourceExistsError("Staff with same email already exists")
+        # Fetch user details from Keycloak
+        users = KeycloakService.get_user_by_email(email)
+        if not users:
+            raise ResourceNotFoundError(f"No user found with email: {email}")
```
dinesh-aot: It would be great to have UnprocessbleEntity error here. We are trying to create a staff here, so it won't be good to have ResourceNotFound.

epictrack-api/src/api/services/staff.py
```diff
@@ -70,9 +80,15 @@ def create_staff(cls, payload: dict):
     @classmethod
     def update_staff(cls, staff_id: int, payload: dict):
         """Update existing staff."""
-        exists = cls.check_existence(payload["email"], staff_id)
+        email = payload["email"].lower()
+        exists = cls.check_existence(email, staff_id)
         if exists:
             raise ResourceExistsError("Staff with same email already exists")
+        users = KeycloakService.get_user_by_email(email)
```
dinesh-aot: I don't think you need to do it in update staff. Because when you created the staff, the result has the idir_user_id. then each time the staff resource should have the idir_user_id associated with it when it is updating. You can make idir_user_id as a required field in update_staff_request object.

PR #2419, opened 2024-10-29
175: Fix permission getter due to other groups coming back from KC.

marklise:


Marilynn-Stone (approved):

epictrack-api/src/api/services/user.py
```diff
@@ -56,7 +56,11 @@ def get_groups(cls):
         filtered_groups = []
 
         for group in groups:
-            current_app.logger.info(f"group: {group}")
+            # For some reason we get all the groups from keycloak instead of just requesting the
+            # TRACK group. So we need to filter out the groups that are not TRACK based, otherwise
+            # we will get all the groups in the system.
+            if group.get("name") != "TRACK":
```
dinesh-aot, line 62: Should this condition be the opposite. We only need to get the sub groups if the main group name is equal to TRACK.

PR #2428, opened 2024-11-10
255: Visual high priority notes for UI. Quality changes for debugging…

marklise:
…. Only deals with UI and not PDF generation - that is to come.

### DIFF
--- a/epictrack-api/src/api/reports/report_factory.py
+++ b/epictrack-api/src/api/reports/report_factory.py
@@ -5,7 +5,7 @@ from collections import defaultdict
 from io import BytesIO
 from pathlib import Path
 from sqlalchemy import and_, func
-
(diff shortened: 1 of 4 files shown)


epictrack-api/src/api/reports/report_factory.py
```diff
@@ -36,7 +36,8 @@ def _format_data(self, data):
             excluded_items = self.filters["exclude"]
         for item in data:
             obj = {
-                k: getattr(item, k) for k in self.data_keys if k not in excluded_items
+              k: (item[k] if k != "work_issues" else self._deserialize_work_issues(item[k]))
```
Ckoelewyn, line 39: 👍I like this

Ckoelewyn (approved):

PR #2459, opened 2024-11-28
[Track-274-2] Fix AS decision maker 

tolkamps1:


tolkamps1:
also Track-223 change order of displayed names to first_name last_name

marklise (approved):

PR #2467, opened 2024-12-05
[Track-297] RF update work title in model to remove hanging dash

tolkamps1:


tolkamps1:
I modified the test because it was testing incorrectly? Please double-check because my brain is mush but the test should assert that ` exists: True ` when trying to create a new work with the title of an existing work.

marklise (approved):

PR #2492, opened 2025-01-03
TRACK-307, add new ministries

tom0827:


epictrack-api/migrations/versions/eccbb8b96011_add_new_ministries.py
```diff
@@ -0,0 +1,32 @@
...
+# revision identifiers, used by Alembic.
+revision = 'eccbb8b96011'
+down_revision = '09fb62ec79e9'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.execute("UPDATE ministries SET sort_order = sort_order + 7 WHERE name != 'Not Applicable'")
+    op.execute("INSERT INTO ministries (name,abbreviation,sort_order) values ('Mining and Critical Minerals','MCM',1)")
```
tolkamps1: Sort order should start at 2. `Not Applicable` is 1.

tolkamps1:
Do we need to deprecate the old ministries at the same time? They all show up now when creating a new work.

tom0827:
Connor said for now we should leave all in so staff can do historical data entry

tolkamps1 (approved):

Marilynn-Stone (approved):

PR #2534, opened 2025-02-19
[Track-26] Active switch needs two clicks

tolkamps1:
- refactor to use `TrackDialog` like other lists
- remove master context


epictrack-web/src/components/indigenousNation/Dialog/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+        String(firstNationId)
+      );
+      setFirstNation(response.data);
+    } catch (error) {
+      showNotification("Could not load First Nation", { type: "error" });
+    }
+  }, [firstNationId]);
+
+  const saveFirstNation = async (data: any) => {
+    console.log("in saveFirstNation");
```
Ckoelewyn: Need to tidy these up champ

Ckoelewyn (changes requested):
All good after those console.logs are removed :)

Ckoelewyn (approved):

PR #2596, opened 2025-05-29
Track 87

Ckoelewyn:


epictrack-api/scripts/refresh-test-db.sh
```diff
@@ -0,0 +1,97 @@
...
+        echo "restoring database from $DUMPFILENAME"
+        psql  -U "$TEST_USER" -p $LOCAL_PORT -h localhost -d postgres -c "CREATE DATABASE app;" || true
+        psql  -U "$TEST_USER" -p $LOCAL_PORT -h localhost -d app -c "CREATE ROLE app;" || true
+        pg_restore -h localhost -U "$TEST_USER" -p $LOCAL_PORT -d app -c ~/"$DUMPFILENAME"
+    fi
+fi
+
+# Cleanup
+unset TEST_PW
+unset $PGPASSWORD
```
tolkamps1: extra `$`

epictrack-api/scripts/refresh-test-db.sh
```diff
@@ -0,0 +1,97 @@
...
+# Step 3: Start port-forwarding in the background
+echo "Port-forwarding database pod..."
+POD_NAME="patroni-epictrack-db-0"
+LOCAL_PORT=15432
+REMOTE_PORT=5432
+
+# Start port-forward in background and save PID to kill later
+oc port-forward "$POD_NAME" ${LOCAL_PORT}:${REMOTE_PORT} &
+PF_PID=$!
+sleep 3  # Give it a moment to establish connection
```
tolkamps1, line 34: Do we want to actually check the connection was successful or print out an error if the dump fails and it could be because the port-forward failed?

tolkamps1 (changes requested):
noice

epictrack-api/scripts/refresh-test-db.sh
Ckoelewyn: If the Port forwards fails, it screams at you. I know because it screamed at me :)

epictrack-api/scripts/refresh-test-db.sh
Ckoelewyn: GONE!

tolkamps1 (approved):

PR #2619, opened 2025-07-07
[Track-196] Delete special history entries

tolkamps1:


epictrack-api/src/api/services/special_field.py
Ckoelewyn, line 1: Would make feel a lot better if we checked the user's role for permissions to delete on the API

tolkamps1: Then you'll have to update your keycloak permissions excel sheet. :P

Good call, ill add that in

Ckoelewyn (approved):

PR #2634, opened 2025-07-23
[TRACK-582] update default filters on work list

Ckoelewyn:


tolkamps1:
remove the `is_active: true`? :P

tolkamps1 (approved):

PR #2674, opened 2025-09-09
[TRACK-96] Add sort order toggle to my workplans page

tom0827:


epictrack-api/src/api/resources/work.py
```diff
@@ -82,6 +82,7 @@ def get():
             work_types=list(map(int, args.getlist('work_types[]'))),
             text=args.get('text', None, str),
             staff_id=args.get('staff_id', None, int),
+            sort_order=args.get('sort_order', 'desc', str),
```
tolkamps1: `sort_order` is already part of the pagination_options (above, line 75). I know the sorting for the work plans wasn't done that way before, but to keep it consistent with the other dashboards, I  think we should keep it under pagination.

tolkamps1 (changes requested):

tolkamps1 (approved):

PR #2684, opened 2025-09-17
[TRACK-456] Responsive insight charts to table filters

tom0827:


epictrack-api/src/api/insights/insights_table_filters.py
tolkamps1, line 1: Is there a cleaner way to do this? Maybe creating a map, or separating filters based on the different tables like the project table and work table? I don't think there is any overlap between the two.

tolkamps1 (approved):

PR #2691, opened 2025-09-24
[Track-634] calendar tasks use correct schema response

tolkamps1:


epictrack-web/src/components/calendar/EventCalendarContext.tsx
```diff
@@ -185,9 +185,8 @@ export const EventCalendarProvider = ({
       const mappedTasks = taskItems.map((element) => ({
         event: {
           ...element.event,
-          start_date: element.start_date,
```
tom0827, line 188: Does this also need to be element.event.start_date or just the one below?

tolkamps1: with the taskeventschema the `start_date` is already passed as part of the event so it doesn't need to be calculated.

tom0827 (approved):

PR #2700, opened 2025-10-03
[TRACK-519] Add staff user filter to phase insight tab

tom0827:


tolkamps1 (commented):
I think the staff_id filter needs to be added to the `useGetWorkPhasesQuery` and the `PhaseInsightsContext` to effect the tables and the filter options

tolkamps1 (approved):

PR #2755, opened 2025-12-10
[TRACK-688-2] fix extension requirement calculation

tom0827:


epictrack-api/src/api/services/event.py
```diff
@@ -665,17 +666,20 @@ def _validate_no_extension_required_to_complete_work(
         throw_error: bool = True,
     ):
         """Validate that overages have been dealt with before allowing the last phase to complete"""
-        total_overage_days = 0
-        for work_phase in all_work_phases[:-1]:  # all except last phase
-            if work_phase.legislated:
-                total_overage_days += work_phase.get_overage_days()
```
tolkamps1, line 671: `get_overage_days()`
Think this might not be used at all anymore. imo we should remove it because phase.end_date has been known to be unreliable regardless and is no used in any other calculations because of that.

tolkamps1 (changes requested):

tolkamps1 (approved):

PR #2827, opened 2026-07-31
feat(styling): align colours and typography with BC Design System (Track-785)

Ckoelewyn:
Colour and type values were spread across a hand-maintained Palette object and 206 hex literals (138 unique) that bypassed the MUI theme entirely. Palette and BaseTheme were two unsynced sources: only neutral/primary/secondary/error reached MUI, so warning, success, info, background, divider, text and action all fell through to MUI's defaults and the 126 files importing Palette directly could not be restyled through the theme at all.

Both now resolve from a single set of design tokens in src/styles/designTokens.ts, transcribed from @bcgov/design-tokens v5.0.0 with the upstream names preserved so the two can be diffed. Kept local rather than taken as a dependency so values can be tuned without waiting on an upstream release. Values the design system does not publish live in a fenced "EPIC.track additions" section, prefixed track*. A small SCSS twin in _tokens.scss serves App.scss and the rich text editor styles.

Hex literals in epictrack-web/src drop from 206 to 141; every survivor is a domain palette (work legend, chart colours, the editor's colour picker), a token definition or a test fixture, documented for follow-up.

Three defects made the previous styling worse than it looked:

- The theme's font stack used U+00B7 MIDDLE DOT instead of spaces as separators, so the whole font-family declaration was invalid and the app rendered by inheritance from body alone. All 11 fontFamily sites in components/shared were no-ops.
- A single @font-face listed Regular and Bold woff/woff2 in one src list with no font-weight descriptor, so the browser treated them as interchangeable sources of one weight and synthesized bold. Replaced with @bcgov/bc-sans 2.1.0 and BC_Sans.css, which declares the "BC Sans" family name the tokens expect and one face per weight; the vendored font binaries and their committed __MACOSX stubs are deleted.
- h1, h2 and h3 had a line-height smaller than their font-size, so headings clipped. Sizes are unchanged; each variant takes the upstream line height paired with its size, and headings are now bold.

Also in this change:

- Fill in the missing palette entries so validation, disabled, placeholder and divider states stop resolving to MUI defaults, and add focus-visible outlines across buttons, icon buttons, checkboxes, radios, tabs, chips and links. The app previously had three :focus rules in total against disableRipple everywhere.
- Stop outlined and text buttons with color="secondary" picking up the contained-secondary rules. The variant branches were joined with || and the secondary branch was gated on colour but not variant, so a disabled outlined-secondary button rendered as a filled grey block.
- Split input hover, focus and error borders, which shared one colour, and add MuiAlert overrides so the three raw MUI alerts on the report screens stop rendering in stock MUI colours.
- Fix form validation that never reached the user: ControlledMultiSelect rendered its helper text only when the caller passed one, so validation messages never showed; ControlledSelectV2 could render the literal string "undefined"; ControlledTextField let a caller silently override validation; checkbox and switch had no error state.
- Drive the ET* typography wrappers from the theme variants and spread caller sx last, so callers can override the weight and size. Nine of ten wrappers spread it first. ETFormLabel now forwards htmlFor, error and sx instead of dropping them, and ETGridTitle keeps its tooltip and ellipsis when disabled.
- Give ETChip's inactive case and StatusBadge the same tokens, so active and inactive stop looking different on My Updates than everywhere else.
- Replace var(--neutral-background-dark, #DBDCDC) in seven places; the custom property is defined nowhere, so every one always used its fallback.
- Size the My Workplans card body to its content and stop it wrapping. MUI Grid containers wrap by default, so on a fixed-height column container the taller text started a second column and the federal involvement value rendered beside the work state badge. CardSkeleton now derives its height from the card instead of a hardcoded 446 that had drifted from the real 451.
- Truncate the Work by Nation axis labels to one line at a wider axis. Nation names run to 52 characters against width={100}, which recharts wrapped to four or five lines inside a 30px row.

Existing behaviour is otherwise unchanged: 518 Cypress tests across 173 specs pass, lint and the production build are clean, and typecheck is unchanged from its 106 pre-existing errors. Manual verification at the supported breakpoints is still outstanding, as is design sign-off on the colour shift and on headings becoming bold and taller.

Ref Track-785


epictrack-web/src/components/shared/controlledInputComponents/ControlledSelectV2.tsx
```diff
@@ -69,7 +69,10 @@ const ControlledSelectV2: React.ForwardRefRenderFunction<
             }}
             disabled={disabled}
             error={!!error}
-            helperText={String(error?.message) || helperText}
+            // `String(undefined)` is the truthy string "undefined", so the old
```
tolkamps1: i think we can remove this comment

epictrack-web/src/components/shared/CustomSwitch.tsx
```diff
@@ -18,17 +19,19 @@ export const CustomSwitch = styled((props: SwitchProps) => (
...
         backgroundColor: `${
           props.disabled ? Palette.neutral.light : Palette.primary.accent.main
         }`,
         opacity: 1,
       },
     },
-    "&.Mui-focusVisible .MuiSwitch-thumb": {
-      color: "#33cf4d",
-      border: "6px solid #fff",
+    // Was an off-palette green thumb; the focus ring now matches every other
```
tolkamps1: I don't think this comment provides any additional insight. commit message addresses this change.

epictrack-web/src/styles/constants.ts
```diff
@@ -1,4 +1,13 @@
...
-export const MET_Header_Font_Weight_Bold = 700;
-export const MET_Header_Font_Weight_Regular = 400;
+import {
+  typographyFontFamiliesBcSans,
+  typographyFontWeightsBold,
+  typographyFontWeightsRegular,
+} from "./designTokens";
+
+// Single source for the font stack. The token supplies its own quotes:
+// "'BC Sans'". This previously used U+00B7 MIDDLE DOT characters instead of
```
tolkamps1: I think comments should pertain to how things are currently done if the process needs explaining. This adds so much detail on what was previously done and that it didn't work, but that is no longer applicable to the app and adds bloat.

epictrack-web/src/index.tsx
```diff
@@ -5,7 +5,9 @@ import { store } from "./store";
 import { Provider } from "react-redux";
 import { StyledEngineProvider, ThemeProvider } from "@mui/material/styles";
 import { BaseTheme } from "./styles/theme";
-// import "@bcgov/bc-sans/css/BCSans.css";
+// BC_Sans.css (not the legacy BCSans.css) declares the "BC Sans" family name the
```
tolkamps1: unnecessary comment imo. maybe review other inline comments as well as lots of them can be briefer/aren't' necessary as the reason is in the commit message/description.

tolkamps1 (changes requested):

tolkamps1 (approved):

